PR #7, opened 2026-06-15
Record/Replay context

sfc-gh-truwase:
Simplify record/replay API


arctic_platform/rl/utils/record_replay.py
```diff
@@ -0,0 +1,130 @@
...
+        return None
+    return Path(gen_dir)
+
+
+def _replay_generation_mode() -> str:
+    mode = os.environ.get("REPLAY_GENERATION_MODE", "record").lower()
+    if mode not in ("record", "load"):
+        raise ValueError(
+            f"Invalid REPLAY_GENERATION_MODE={mode!r}; expected 'record' or 'load'"
+        )
```
sfc-gh-sbekman: I find the separation between record and replay to be an unnecessary issue which could waste a lot of time, since one needs to remember to switch modes multiple times, rather than never.

I think a much smoother experience is a on/off mode, so all you need to specify is the env var with the path and dynamically choose to replay if the record is there and record if it's not there.

Perhaps we could support both approaches, with the default behavior being just on/off - so if REPLAY_GENERATION_MODE is not set, but REPLAY_GENERATION_DIR is set it automatically records/replays based on the file's existence.

I do see the value in separating things, so ideally we would support both.

arctic_platform/rl/utils/record_replay.py
```diff
@@ -0,0 +1,130 @@
...
+            return torch.load(replay_path, weights_only=False)
+
+
+class RecordReplayContext:
+    """Per-record ``with`` wrapper around :class:`RecordReplay`.
+
+        with record_replay.generation(c) as gen:
+            if gen.output is None:
+                gen.output = generate_sequences(...)
+            gen_batch_output = gen.output
```
sfc-gh-sbekman: Does this read better?

```suggestion
        with record_replay.generation(c) as gen:
            gen_batch_output = gen.output or generate_sequences(...)
```

arctic_platform/rl/utils/record_replay.py
```diff
@@ -0,0 +1,128 @@
...
+                )
+            pr0(f"[REPLAY_GENERATION] loading {replay_path}")
+            return torch.load(replay_path, weights_only=False)
+
+
+class RecordReplayContext:
+    """Per-record ``with`` wrapper around :class:`RecordReplay`.
+
+        with record_replay.generation(c) as gen:
+            gen_batch_output = gen.output or generate_sequences(...)
```
sfc-gh-truwase: It just occurred to me that we won't be able to save `gen_batch_output` in record mode. Or am I missing something?

```suggestion
        with record_replay.generation(c) as gen:
            gen_batch_output = gen.output or generate_sequences(...)
            gen.output = gen_batch_output # do we need something like this?
```

sfc-gh-sbekman: you're correct, Tunji

OK, how about this then. Instead of a context manager we use a decorator - that way you don't need to change the caller at all - just 

```
@generate_maybe_replay()
def generate_sequences
```

sfc-gh-kganesan (approved):

PR #8, opened 2026-06-15
Fix rollout-compile grad-norm explosion in colocate by offloading wei…

xylian86:
# Fix rollout-compile issue that causes grad-norm explosion 

## Summary

In colocate mode the per-step weight offload/backload reallocated `param.data` on every wake, **changing each weight's GPU virtual address**. With rollout compile enabled (`enforce_eager=False`), vLLM's captured CUDA graph keeps reading the *old* addresses, so generation samples from stale/garbage memory. 

Fix: stop the legacy manual offload (`offload_weights=False`) and let vLLM's `CuMemAllocator` free the weights. cumem returns only the physical pages and keeps the virtual address pinned.

### DIFF
--- a/arctic_platform/rl/http_server.py
+++ b/arctic_platform/rl/http_server.py
@@ -511,10 +511,15 @@ async def sleep_inference(job_id: int, level: int):
     colocate = app.state.colocate
     results = {}
     pool: ReplicaPool = app.state.sampling_pool
-    results["sampling"] = await pool.sleep(level=level, offload_weights=colocate)
+    # Let vLLM's CuMemAllocator free the weights (offload_weights=False) instead
+    # of the legacy manual offload, which reallocated param.data on each wake and
+    # changed weight addresses -> stale rollout CUDA graphs (compile on) ->
+    # grad-norm explosion. cumem keeps addresses stable.
+    offload_weights = False
+    results["sampling"] = await pool.sleep(level=level, offload_weights=offload_weights)
     lp_pool: ReplicaPool | None = app.state.log_prob_pool
     if lp_pool is not None and lp_pool._config is not None and not lp_pool.sleeping:
-        results["log_prob"] = await lp_pool.sleep(level=level, offload_weights=colocate)
+        results["log_prob"] = await lp_pool.sleep(level=level, offload_weights=offload_weights)
     if colocate:
         await pool.close_weight_sync()
         if lp_pool is not None and lp_pool._config is not None:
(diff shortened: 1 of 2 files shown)


arctic_platform/rl/http_server.py
```diff
@@ -511,10 +511,15 @@ async def sleep_inference(job_id: int, level: int):
     colocate = app.state.colocate
     results = {}
     pool: ReplicaPool = app.state.sampling_pool
-    results["sampling"] = await pool.sleep(level=level, offload_weights=colocate)
+    # Let vLLM's CuMemAllocator free the weights (offload_weights=False) instead
+    # of the legacy manual offload, which reallocated param.data on each wake and
+    # changed weight addresses -> stale rollout CUDA graphs (compile on) ->
+    # grad-norm explosion. cumem keeps addresses stable.
+    offload_weights = False
```
sfc-gh-truwase, line 518: While I see the conflict of `offload_weights` and `cuda_graph`, I recall that in some colocate situations, `offload_weights` is needed to avoid OOMs because cuMemAllocator` still holds on to some memory.

sfc-gh-truwase, line 518: Is it possible to disable conditionally when cuda_graph/compile is enabled? In other words, something like 
```python 
    offload_weights = colocate and not cuda_graph`
```

arctic_platform/rl/ray_server.py
```diff
@@ -253,7 +253,11 @@ async def sleep_inference(self, job_id: int, level: int) -> dict[str, Any]:
         colocate = self.colocate
         results = {}
         pool: ReplicaPool = self.sampling_pool
-        offload_weights = colocate
+        # Let vLLM's CuMemAllocator free the weights (offload_weights=False)
+        # instead of the legacy manual offload, which reallocated param.data on
+        # each wake and changed weight addresses -> stale rollout CUDA graphs
+        # (compile on) -> grad-norm explosion. cumem keeps addresses stable.
+        offload_weights = False
```
sfc-gh-truwase, line 260: Same here

arctic_platform/rl/http_server.py
xylian86: What's retained after sleep is the virtual address reservation, not physical VRAM — unmap_and_release() returns the physical pages, and virtual address space doesn't count against GPU memory. And this is also the exact same cumem sleep/wake that verl's baseline rollout uses (engine.sleep/wake_up, no manual offload); after this PR the colocate path just relies on that same path, so the offload is now identical to the verl baseline.

xylian86: Possible, but I'd suggest offload_weights=False unconditionally rather than gate on cuda_graph:
- No memory upside: collective_rpc("sleep") runs either way and CuMemAllocator
  frees the same physical pages. The manual path only adds D2H/H2D + param.data realloc.
- Matches baseline: verl baseline uses cumem sleep/wake for both eager and
  compile, never a manual offload — so False everywhere == baseline.

sfc-gh-truwase (approved):

PR #31, opened 2026-06-29
verl recipes

sfc-gh-truwase:
- baseline scripts
- simplify


sfc-gh-sbekman (approved):
thank you for the improvements, Tunji

PR #43, opened 2026-07-23
feat(rl): add Tinker HTTP layer for Arctic-Platform (v1)

sfc-gh-kganesan:
## Summary

Adds a Tinker-compatible HTTP layer to the Arctic-Platform RL server so any upstream `tinker.ServiceClient` can drive Arctic without knowing the native `/fwd-bwd` / `/step` / `/generate` shape. New router is mounted at `/api/v1/*`, lives alongside the existing native routes in the same FastAPI process, and lowers into Arctic handlers via in-process calls (no second HTTP hop).

**Scope (v1):** RL only, colocated (`colocate=True`, CUDA-IPC weight sync), single global training run, no auth. Full-weight DeepSpeed training via the SkyRL-tx `LoraConfig(rank=0)` = FFT convention; `rank>0` returns HTTP 400. LoRA + async-RL are captured as extensions `E1..E6` in the design doc.

Full design doc: [`arctic_platform/rl/TINKER_COMPAT.md`](arctic_platform/rl/TINKER_COMPAT.md).

## Wire protocol coverage

All wire routes are `POST /api/v1/<verb>`. Every long-running verb returns a future ([`UntypedAPIFuture`](https://github.com/thinking-machines-lab/tinker/blob/main/src/tinker/types/shared/untyped_api_future.py)); v1 executes inline and stashes the terminal response so `retrieve_future` returns the result on the first poll.

| Route | Arctic handler | Status |
| :--- | :--- | :--- |
| `create_session`, `session_heartbeat`, `client/config`, `auth/token`, `telemetry` | in-memory / no-op | done |
| `get_server_capabilities` | static (server's own `base_model`) | done |
| `create_model` | full-weight if `LoraConfig(rank==0)`, else 400 | done |
| `get_info` | in-memory model registry | done |
| `create_sampling_session` | binds to current weight generation | done |
| `forward` | `fwd_bwd(..., forward_only=True)` | done |
| `forward_backward` | `fwd_bwd(...)` with `loss_fn_config` threaded into Arctic `actor_config` | done |
| `optim_step` | `step(optim_overrides=...)` (new param, this PR) | done |
| `save_weights_for_sampler` | `sync_weights(cuda_ipc=True)` + bump weight-gen | done |
| `asample` | `generate(...)` gated on `sampling_session_id` staleness | done |
| `retrieve_future` | pop from in-memory store; `TryAgainResponse` if unknown | done |
| `load_weights`, `save_weights`, `unload_model`, `weights_info`, `training_runs/*` | — | out of scope (E1..E3) |

Loss functions supported in v1: `ppo`, `importance_sampling`. `cispo` / `dro` / `cross_entropy` return HTTP 400 (documented; `cross_entropy` lands in `E3`).

## Changes

**Plugin-side (this PR):**

- `arctic_platform/rl/tinker_server.py` (new, ~1000 LoC) — FastAPI router, Pydantic wire models mirroring `tinker.types.*`, adapters, in-memory future store, weight-generation counter.
- `arctic_platform/rl/deepspeed_worker.py` — `step()` gains optional `optim_overrides` dict (lr / betas / eps / weight_decay) applied to `param_groups` before `engine.step()`. Legacy no-override path unchanged.
- `arctic_platform/rl/http_server.py` — `/step` accepts `StepRequest{optim_overrides}`; tinker router is mounted eagerly.
- `arctic_platform/rl/ray_server.py` — matching passthrough for the Ray transport.
- `arctic_platform/rl/utils/server_models.py` — adds `StepRequest`.
- `arctic_platform/rl/__init__.py` — PEP 562 lazy imports so `tinker_server` can be imported in a CPU-only env (needed to keep tests lightweight; existing eager consumers keep the same public surface).
- `arctic_platform/rl/TINKER_COMPAT.md` (new) — full design doc: verb map, RL step data flow, futures model, sync vs async semantics, concrete code changes, tests, extensions (`E-async`, `E1..E6`), and test results.

**Net diff:** +3195 / -31 lines across 14 files.

## Tests

All 75 tests pass in ~2s on a CPU-only conda env (`/data-fast/karthik/conda_envs/tinker_dev`, Python 3.12, `tinker` SDK 0.x, no torch / no vLLM / no DeepSpeed).

\`\`\`
$ python -m pytest tests/tinker_layer/ -v
============================= 75 passed in 2.03s ==============================
\`\`\`

| Module | Tests | What it covers |
| :--- | :---: | :--- |
| \`test_wire_schema.py\` | 21 | request/response Pydantic models parse valid upstream bodies and reject malformed ones; \`TensorData\` sparse encoding round-trips; \`ModelInput\` rejects non-\`encoded_text\` chunks. |
| \`test_adapters.py\` | 22 | \`datum_list_to_arctic_batch\` shape + **ZoRRo padding invariant** (config maxes, not batch-local); \`loss_fn_config\` → Arctic \`actor_config\` (\`ppo\` clip low/high, \`importance_sampling\` disables clip, \`kl_coef\`, \`entropy_coef\`); \`AdamParams\` / \`SamplingParams\` round-trips (stop as str / list[str] / list[int]). |
| \`test_tinker_server.py\` | 27 | one test per HTTP verb via \`httpx.AsyncClient\` in-process, against a mocked Arctic backend: bootstrap, model lifecycle (including \`rank>0\` → 400 and wrong \`base_model\` → 400), training (unsupported loss matrix), weight sync (bump gen), sampling (current gen serves, stale → 409, no-session serves), futures (unknown → try_again, pop semantics), misconfig (unwired app → 500). |
| \`test_rl_loop.py\` | 5 | **acceptance test** — uses the real upstream \`tinker.ServiceClient\` against a \`uvicorn\`-hosted app; \`ServiceClient\` bootstrap, \`create_lora_training_client(rank=0)\` accepted, \`rank>0\` raises, a single RL step, and a full 10-step RL loop (\`save_weights_and_get_sampling_client\` → \`sample\` → \`forward_backward\` → \`optim_step\`). Proves wire compatibility with the upstream SDK end-to-end. |

### Interop notes surfaced by the test suite

Called out here so a reviewer can spot-check the assumptions (all fixed in this PR):

1. **Metric names carry \`:reduction\` suffix.** Tinker's \`combine_fwd_bwd_output_results\` splits every metric key on \`:\` to look up the reduction function; plain \`\"loss\"\` from Arctic must be annotated as \`\"loss:mean\"\`. See \`arctic_metrics_to_tinker\`.
2. **\`ForwardBackwardOutput.loss_fn_outputs\` doubles as the reduction weight.** \`len(loss_fn_outputs)\` is the per-actor weight for \`:mean\` / \`:slack\`; empty list → \`ZeroDivisionError\` in \`np.average\`. Arctic returns one aggregated batch, so the adapter emits one empty \`LossFnOutput\` per input \`Datum\`.
3. **\`TelemetryResponse\` expects \`status: Literal[\"accepted\"]\`,** not \`received: bool\`. Pinned via the SDK's pydantic model.

## Test plan

- [x] \`pytest tests/tinker_layer/\` (75 tests, CPU-only, ~2s).
- [ ] Point a real \`tinker.ServiceClient\` at an Arctic http_server on a full GPU box and run a small RL loop end-to-end (blocked on a separate branch since it needs the full \`arctic_verl_v18\` env; already covered by the CPU-only 10-step conformance test).
- [ ] Follow-up PR: land \`E-async\` (real async future execution) so \`forward_backward\` + \`save_weights_for_sampler\` + \`asample\` can pipeline instead of serializing behind each inline call.

Co-authored-by: Cursor <cursoragent@cursor.com>


sfc-gh-kganesan:
E2E driven by the [SkyRL-tx cookbook `rl_loop.py`](https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/recipes/rl_loop.py) recipe (the reference RL loop advertised by the [SkyRL-tx README](https://github.com/NovaSky-AI/SkyRL/tree/main/skyrl-tx)) against Arctic — no cookbook changes:

```bash
python -m arctic_platform.rl.http_server \
    --host 0.0.0.0 --port 7000 \
    --training-gpus 1 --sampling-gpus 1 --colocate \
    --base-model Qwen/Qwen3-1.7B \
    --tinker-max-prompt-length 1024 --tinker-max-response-length 512

TINKER_API_KEY=tml-dummy python -m tinker_cookbook.recipes.rl_loop \
    base_url=http://localhost:7000 model_name=Qwen/Qwen3-1.7B lora_rank=0 \
    batch_size=2 group_size=4 max_steps=5 max_tokens=32 \
    save_every=0 learning_rate=1e-6
```

5 optim steps, gsm8k prompts, cookbook-issued `forward_backward(loss_fn="importance_sampling")` + `optim_step` + `save_weights_and_get_sampling_client`:

| step | reward/total | grad_norm:mean | time/total (s) |
| ---: | ---: | ---: | ---: |
| 0 | 0.250 | 2.92 | 1.50 |
| 1 | 0.250 | 1.93 | 0.57 |
| 2 | 0.125 | 3.49 | 0.48 |
| 3 | 0.125 | 3.71 | 0.45 |
| 4 | 0.375 | 3.06 | 0.55 |

Loop exits cleanly with `Training completed` and writes a final checkpoint (sampler snapshot + `save_state` ack).

Follow-on changes pushed to `karthik/tinker-http-layer` in this run:
- adapter: infer prompt/response boundary from `advantages` / `target_tokens` when `weights` isn't set — the shape the SkyRL-tx cookbook uses.
- server: ack-only `/api/v1/save_weights` so cookbook recipes ending in `save_checkpoint(kind="both")` don't 404 on shutdown (real state persistence remains E2).
- doc: trimmed `TINKER_COMPAT.md` from ~820 → ~440 LoC, added this validation table.
- tests: `pytest tests/tinker_layer/` → 76 passed.

sfc-gh-kganesan:
@tjruwase — would love your read on the DeepSpeed side of this change. Question is **whether `arctic_platform/rl/deepspeed_worker.py` should be touched at all** for Tinker support, and if so, whether there's a cleaner shape than what's in `7e22756`.

## The mismatch

Tinker's HTTP protocol decouples `/forward_backward` from `/optim_step` — client-driven boundary. The upstream `tinker` SDK also chunks each logical training step into multiple ~5 MB HTTP calls (`MAX_CHUNK_BYTES_COUNT` / `MAX_CHUNK_LEN=1024` in `training_client.py`), so one training step arrives at the server as N `/forward_backward` calls followed by exactly one `/optim_step`.

DeepSpeed has an **implicit** grad-accum boundary driven by an internal counter (`deepspeed/runtime/engine.py:is_gradient_accumulation_boundary`):

```python
return (self.micro_steps + 1) % self.gradient_accumulation_steps() == 0
```

Every `engine.backward()` bumps `micro_steps`; every `engine.step()` checks that predicate and silently runs the optimizer + zeros grads when it's true. FSDP-based backends (SkyRL-tx) sidestep this entirely — they call `optimizer.step()` when they want. DeepSpeed's only override is `engine.set_gradient_accumulation_boundary(bool)`, an instance method that only exists on the engine inside the Ray actor.

## What the diff does (~15 lines guarded by `meta["tinker_grad_accum"]`)

1. **Different micro-batch split.** Arctic's original path assumes `micro_batches == gradient_accumulation_steps`. Tinker's chunks can be smaller than `grad_accum_steps` (e.g. `grad_accum_steps=64`, chunk shard on this DP worker=47), so `split_dict(chunk, grad_accum_steps)` blows up with `Batch dimension 47 < num_workers 64`. Tinker mode splits by `train_micro_batch_size_per_gpu` instead.
2. **Skip interior `engine.step()`.** The original path calls `engine.step()` between micro-batches to advance `micro_steps`; in Tinker mode this would trip the implicit boundary mid-chunk.
3. **Explicit boundary control.** `set_gradient_accumulation_boundary(False)` per micro-backward; `.step()` (called by `/optim_step`) flips it back to `True` and does exactly one "real" engine step so grads apply + zero.

Verl callers hit zero of this (flag defaults to `False`) — the count-based path is unchanged.

## Alternatives I rejected

1. **Buffer chunks in the adapter, replay on `/optim_step`.** Breaks Tinker's per-call response contract: `ForwardBackwardOutput.loss_fn_outputs[i].logprobs` has to come back on the same call the SDK sent (cookbook's `compute_kl_sample_train` indexes it before the next optim step). Also holds the full ~2 GB step batch in adapter-process memory.
2. **`gradient_accumulation_steps=1` at init time.** Applies the optimizer after every chunk → N tiny SGD-style updates per logical step → wrong optimization dynamics.
3. **Pre-configure `grad_accum_steps` = SDK chunk count.** Chunk count is client-controlled and variable (batch bytes / `MAX_CHUNK_BYTES_COUNT`); the server can't know it up front.
4. **New `/fwd-bwd-accumulate` HTTP endpoint that skips step.** Same underlying worker change, worse coupling (public endpoint for what's really an internal flag).

## Where I'd love your input

- **Am I misreading DS semantics?** Specifically, is there a way to get "accumulate across N `.backward()` calls, apply exactly once on an external trigger" *without* toggling `set_gradient_accumulation_boundary` per backward? A DS-idiomatic pattern I'm missing?
- **Is the meta-flag naming right?** `tinker_grad_accum` reads as vendor-specific. `defer_optim_step` or `external_grad_accum_boundary` might be more honest about intent.
- **Is there a first-class DS API for this use case?** The "backward and step are separate primitives, client owns the boundary" pattern is going to hit any DS-backed OSS RL backend that wants to speak this protocol. If there's an upstream ask that would make this cleaner, worth raising it.
- **Can this shrink further?** I could collapse the `if/else` split-selection into `n_splits = tinker_mode ? ceil(batch/micro_bs) : grad_accum_steps` (trims ~4 lines) but it changes the verl control flow, so I kept them separate to keep the risk surface at zero for existing callers. Preference?

Files: [`arctic_platform/rl/deepspeed_worker.py`](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/43/files#diff-arctic_platform-rl-deepspeed_worker-py) — full context in commit `7e22756`.

sfc-gh-truwase:
@sfc-gh-kganesan, I am tracking with https://github.com/deepspeedai/DeepSpeed/issues/8183

sfc-gh-kganesan:
Superseded by #110, which takes the router from this branch and re-homes it. Recording what carried over and what did not, since the delta is the interesting part.

**The router here is already a pure protocol adapter** — it speaks Tinker's HTTP surface and calls five injected handlers, and knows nothing about a backend. That turned out to be the load-bearing design decision in this PR, because it means the on-prem wiring is optional rather than required. #110 moves the file to `arctic_platform/integrations/tinker/router.py` and binds those five handlers to `AsyncArcticRLClient` instead, which reaches Cortex Training without the on-prem server, DeepSpeed, or Ray being involved.

So the changes here to `http_server.py`, `deepspeed_worker.py`, `ray_server.py`, `rl/__init__.py` and `utils/server_models.py` are all dropped. The diff goes from touching the server internals to adding one self-contained subpackage.

**One thing this branch got wrong, worth flagging for its own sake:** the wire is protobuf, not JSON. `tinker==0.25.0` posts `forward_backward` as `application/x-protobuf` and *rejects* a JSON reply for `ForwardBackwardOutput` and `SampleResponse`:

```
Server returned a JSON payload for {model_cls}, which this SDK version only
supports as proto. The server predates proto response serialization for this type.
```

A JSON-only server therefore fails at the first training step regardless of how correct its numbers are. This branch's schema tests pass anyway because they compare the router against hand-copied copies of the SDK's types rather than against the SDK — so they cannot detect upstream drift, which is exactly the failure that occurred. #110 adds a codec and tests it against the SDK's own `request_conv` / `response_conv` converters as the oracle in both directions.

Two other things found while getting it running end to end, both of which affect the datum conversion in this branch:

- `target_tokens[-1]` is a token that never appears in `model_input`, so the last scored position has nothing to predict while the advantage sitting there still multiplies its gradient. It showed up only as `kl_sample_train_v1` reading 0.35–0.49; 0.001–0.005 once the token is appended.
- Cortex's packer needs real tokens in the leading columns, and the log-probs come back in that aligned frame — so the permutation has to be inverted on the way back, or every row's log-probs shift by its own prompt padding with no error raised.

For the record on the end state: an unmodified `tinker_cookbook.recipes.math_rl.train` now completes a full 935-step GSM8K epoch on Cortex, 63% → 74% correct.

Suggest closing this in favour of #110 unless the on-prem binding is wanted separately — the router in #110 is backend-agnostic, so an on-prem handler set could be added there without reopening this.

PR #36, opened 2026-07-10
feat(integrations): land verl RemoteBackend adapter under arctic_platform/integrations/verl/

sfc-gh-kganesan:
Closes #35.

Ships the ArcticRL ⇄ verl adapter as `arctic_platform/integrations/verl/`, loaded by verl at bootstrap via `VERL_USE_EXTERNAL_MODULES`. verl core keeps its generic `RemoteBackend` ABC + registry; all Arctic runtime code moves out of the verl source tree.

## What lands

New subpackage `arctic_platform.integrations.verl/`:

| file | role |
|---|---|
| `register.py` | Loaded by `VERL_USE_EXTERNAL_MODULES`. Eager `import adapter` fires `@RemoteBackendRegistry.register("arctic")`; lazy loaders for the actor-rollout worker and the rollout replica. |
| `adapter.py` | `ArcticRLClientWrapper` — verl's `RemoteBackend` ABC on top of `arctic_platform.rl`. |
| `worker.py` | `ArcticRLActorRolloutRefWorker` — CPU-only forwarder for `ActorRollout(Ref)`. |
| `rollout.py` | `ArcticReplica` / `ArcticLLMServer` — Arctic's vLLM engine as a verl `RolloutReplica`. |
| `grpo_loss.py` | Verl-shaped GRPO loss (`LOSS_FNS["verl_grpo"]`). |
| `config/remote_backend/arctic.yaml` | Per-backend Hydra config. |
| `examples/run_{gsm8k,bird}_grpo_arl.sh` | Single-GPU reference launchers. |

`arctic_platform/rl/processors/verl_grpo.py` becomes a one-line re-export shim so the old import path still resolves. `pyproject.toml` gains a `[verl]` extra.

## Recipe alignment

Both `recipes/rl/verl/{simple,txt2sql}/run_*.sh` pick up two lines each — `export VERL_USE_EXTERNAL_MODULES=…` and `hydra.searchpath=[file://…/integrations/verl/config]`. No hyperparameter changes.

## Tests

`tests/integrations/verl/` — 17 tests, all green on machines without vLLM / DeepSpeed / arctic_inference (`conftest.py` stubs `verl.*`, `arctic_platform.rl.*` before importing the plugin). Covers decorator + lazy registration, `ArcticRLClientWrapper` API + `destroy()` idempotency, `_prepare_padded_arctic_batch_dict` invariants, and the backward-compat shim. Pre-commit clean (`black`, `isort`, `flake8`, `check_license.py`).

[...]

PR #32, opened 2026-06-30
recipes/rl/skyrl: Arctic RL × SkyRL recipes (simple / txt2sql / long_context_qa)

sfc-gh-kganesan:
## Summary

Adds the SkyRL counterpart to `recipes/rl/verl/` — three self-contained Arctic RL × SkyRL recipes that drive SkyRL's GRPO trainer with the Arctic RL backend (ZoRRo train + Forest Cascade Attention + CUDA-IPC weight sync + Liger + FA3 trainer / FLASH_ATTN inference), plus a vendored copy of the SkyRL-side integration code so users don't need a SkyRL checkout.

```
recipes/rl/skyrl/
├── README.md                     # landing page
├── _lib/arctic_rl/               # vendored from SkyRL integrations/arctic_rl @ PR #1837
│   ├── VENDOR.md                 # exact upstream SHA + re-sync instructions
│   ├── {config,trainer,generator,entrypoint}.py
│   ├── __init__.py
│   └── envs/                     # skyrl_gym env registry: bird + long_context_qa
├── simple/                       # GSM8K · Qwen3-0.6B · 1 GPU
├── txt2sql/                      # BIRD-SQL · Qwen3-8B · 1 node × 8 GPUs
└── long_context_qa/              # LoongRL (16K) · Qwen3-8B · 1 node × 8 GPUs
```

Each recipe directory is the same shape as the verl twins: `requirements.txt`, `overrides.txt`, `download_data.py`, `run_*_arl.sh`, `README.md`. No SkyRL clone needed — SkyRL is pulled via `pip install git+...@<commit>` and the recipe-side integration is the vendored `_lib/arctic_rl/`.

## Recipes

| Recipe | Dataset | Model | Topology | Notes |
| --- | --- | --- | --- | --- |
| `simple/`           | GSM8K (HF)         | Qwen3-0.6B | 1 GPU                      | Smallest E2E loop; no Ray / hostfile |
| `txt2sql/`          | BIRD-SQL           | Qwen3-8B   | 1 node × 8 GPUs, TP=2      | Single-node mirror of the multi-node 32B blog run |
| `long_context_qa/`  | LoongRL (16K)      | Qwen3-8B   | 1 node × 8 GPUs, TP=4      | Single-node mirror of the multi-node Qwen3-32B + YaRN-128K blog run |

## Test plan

Smoke-tested E2E on this host (8x H200, conda `skyrl_v2`, branch HEAD at `39eafae`):

[...]

PR #45, opened 2026-07-27
feat(client): unified ArcticRLClient + on-prem Ray transport

sfc-gh-mwyatt:
Add the unified RL client frontend: a single op API (client.py) that lowers every op to a canonical Request(op, job_id, body), a Transport ABC with JobHandles/Request (transport.py), a backend config (config.py), and the shared OnPremTransport base with the in-process Ray transport (onprem.py, onprem_ray.py). ray_server.py accepts (job_id, body) on step/save_checkpoint/reset_prefix_cache so the client can forward uniformly.

Includes tests/client/test_client_ops.py covering the op->Request mapping via a FakeTransport (no GPUs/backends needed).

### DIFF
--- a/arctic_platform/client/config.py
+++ b/arctic_platform/client/config.py
@@ -1,0 +1,58 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""One config for every backend. Switching deployment == changing `backend`."""
+
+from __future__ import annotations
+
+from typing import Any
+from typing import Literal
+
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
+
+JobId = int | str
+
+
+class ArcticRLClientConfig(BaseModel):
+    model_config = ConfigDict(extra="forbid", validate_default=True)
+
+    backend: Literal["onprem"] = Field("onprem", description="Deployment target.")
+    comm_protocol: Literal["ray"] = Field("ray", description="onprem transport: in-process Ray.")
+    model_name: str = Field(..., description="HF model id to train/serve.")
+    seed: int | None = Field(None, description="Global seed.")
+    dtype: str | None = Field(None, description="Parameter dtype override.")
+    max_seq_len: int = Field(8192, description="Max sequence length.")
+
+    # GPU allocation — a job type is created only when its count is > 0.
+    training_gpus: int = 0
+    sampling_gpus: int = 0
+    log_prob_gpus: int = 0
+
+    # onprem (Ray)
+    colocate: bool = False
+    ds_config: dict[str, Any] | None = None
+    training_config: dict[str, Any] | None = None
+    vllm_config: dict[str, Any] | None = None
+    checkpoint_path: str | None = Field(None, description="onprem: dir for training checkpoints/weight sync.")
+
+    # Reconnect: attach to pre-existing jobs instead of creating new ones.
+    training_job_id: JobId | None = None
+    sampling_job_id: JobId | None = None
+    log_prob_job_id: JobId | None = None
+
+    def gpus_for(self, job_type: str) -> int:
+        """GPU count allocated to a job type (0 == the job type is disabled)."""
+        return getattr(self, f"{job_type}_gpus")
--- a/arctic_platform/client/config.py
+++ b/arctic_platform/client/config.py
@@ -46,7 +46,14 @@ class ArcticRLClientConfig(BaseModel):
     ds_config: dict[str, Any] | None = None
     training_config: dict[str, Any] | None = None
     vllm_config: dict[str, Any] | None = None
-    checkpoint_path: str | None = Field(None, description="onprem: dir for training checkpoints/weight sync.")
+    checkpoint_path: str | None = Field(
+        None,
+        description=(
+            "onprem: training job's base checkpoint dir, set at init (resume-from + weight sync). "
+            "This is the default destination; a per-call save_checkpoint(path=...) overrides it for "
+            "that call and falls back to this dir when path is None."
+        ),
+    )
 
     # Reconnect: attach to pre-existing jobs instead of creating new ones.
     training_job_id: JobId | None = None
--- a/arctic_platform/client/examples/sft_example.py
+++ b/arctic_platform/client/examples/sft_example.py
@@ -13,9 +13,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Unified training example for the on-prem Ray backend.
+"""Unified SFT training example for the on-prem Ray backend.
 
-    python arctic_platform/client/examples/train_example.py --backend onprem-ray
+    python arctic_platform/client/examples/sft_example.py --backend onprem-ray
 
 The backend follows the pathway: build config -> ArcticRLClient ->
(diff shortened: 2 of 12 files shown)


sfc-gh-kganesan:
Reviewed the whole three-PR stack end-to-end; full write-up on #47. Notes specific to this PR below.

Ran the tests on this branch: **13/13 pass**.

## Design

The three-layer split — client owns op vocab + job resolution, `Transport` ABC delivers `Request(op, job_id, body, binary)`, `OnPremTransport` base owns creation/ordering/teardown, concretes implement only `_start / _rpc / _destroy / _wait_running` — is exactly the seam that lets #46 (HTTP) and #47 (Cortex) drop in without cross-cutting changes. `JOB_CREATE_ORDER = ("sampling", "log_prob", "training")` with training last as NCCL root is correct and documented. `JobHandles.require()` guard is a nice defensive fence. Docstrings across `transport.py` / `onprem.py` / `client.py` clearly state design intent — reviewer-friendly.

## Plan-level work this PR opens (details on #47 as P3 / P4)

Merging this stack adds a *second* client with a different calling convention rather than replacing the first. AST scan:

```
arctic_platform/rl/http_client.py (ArcticRLHTTPClient) — every op is `async def`
arctic_platform/client/client.py  (ArcticRLClient)     — every op is `sync`
```

Every existing in-repo caller (the `integrations/verl/` `RemoteBackend`, the `recipes/rl/skyrl/` recipes) has to be rewritten before `arctic_platform/rl/client.py` can be deleted. Please add to `UNIFICATION_NOTES.md`: which caller migrates on what PR, when `rl/*_client.py` gets removed. Otherwise "one client" becomes "one-plus-legacy" indefinitely.

Related: ops present in `arctic_platform/rl/*_client.py` but missing from `ArcticRLClient`: `sleep_/wake_{inference,training,log_prob}`, `empty_training_cache`, `weight_norm`. Not on the immediate critical path but they block deleting the async clients. Enumerate in `UNIFICATION_NOTES.md` under a "not yet in unified client" section.

## Branch hygiene

`main` has moved past `#36` (the verl integration adapter); all three PRs in the stack are behind. This branch also carries stray commit `e697039` (`recipes/rl/skyrl … (#32)`) which isn't in `main` and inflates the client-only diff (~866 LOC advertised, +4468 LOC actually on the branch). Rebase before merge, or land `#32` first.

## Not blocking

- `RayTransport._rpc` creates a new event loop per op — `asyncio.run(method(*args))` at `arctic_platform/client/transports/onprem_ray.py:65`. Every `fwd_bwd / generate / step / log_probs` spins up and tears down a fresh loop; on the hot path this is real overhead. Fix: hold one loop on the transport.
- `OnPremTransport.shutdown()` iterates `JOB_TYPES` in insertion order rather than reverse creation order. Usually you want to destroy training before its samplers.
- No `OnPremTransport` unit tests — everything is exercised through `FakeTransport`. A test that pins `_init_payload` composition, `initialize()` ordering, and `shutdown()` traversal would prevent regressions in #46/#47.
- No op-registry validation — if a client op is renamed and a transport misses the update, `AttributeError` at runtime instead of import time.
- `ArcticRLClientConfig.checkpoint_path` overlaps with `save_checkpoint(path=...)`; docstring should clarify which wins when both are set.

None of these gate merge.

arctic_platform/client/transports/onprem.py
```diff
@@ -0,0 +1,98 @@
...
+        # The client already resolved the job id onto the request; just deliver it.
+        return self._rpc(request)
+
+    def shutdown(self) -> None:
+        for job_type in JOB_TYPES:
+            job_id = getattr(self.jobs, job_type)
+            if job_id is not None:
+                self._destroy(job_id, job_type)
+
+    def _init_payload(self, job_type: str) -> dict[str, Any]:
```
sfc-gh-truwase, line 78: This does not appear to match the logic/features [here](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e697039ad921e2bcd2f8da69043cbb1f2050a50e/arctic_platform/rl/ray_client.py#L144), such as 
`self.config.arctic_inference_config`, `use_deepspeed`, `self.config.log_prob_ds_config`, etc. Is that intentional to future work?

sfc-gh-truwase:
@sfc-gh-mwyatt, I think the `gsm8k` recipe of either [verl](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/main/recipes/rl/verl/simple) or [skyrl](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/main/recipes/rl/skyrl/simple_gsm8k) would be good gating tests for these PRs.

arctic_platform/client/examples/sft_example.py
```diff
@@ -0,0 +1,176 @@
...
+
+    with contextlib.ExitStack() as stack:
+        config = profile.config(stack)
+        batch = profile.batch(_tokens())  # same tokens, backend-specific wire shape
+
+        client = ArcticRLClient(config)
+        print(f"training job: {client.jobs.training}")
+        try:
+            for step in range(STEPS):
+                # RL note: a full loop would create a sampling job, generate() a
```
sfc-gh-truwase, line 165: It might be worth renaming this as `sft_example.py` and creating a separate `[grpo|ppo]_example.py` in future.

arctic_platform/client/examples/sft_example.py
```diff
@@ -0,0 +1,176 @@
...
+# Run-from-anywhere: the package root + the RL test harness on the path.
+sys.path.insert(0, str(ARCTIC))
+sys.path.insert(0, str(ARCTIC / "tests" / "rl"))
+
+from arctic_platform.client import ArcticRLClient  # noqa: E402
+from arctic_platform.client import ArcticRLClientConfig  # noqa: E402
+
+STEPS = 20
+SEED = 42
+N_GPUS = 8
```
sfc-gh-truwase, line 46: 2-gpus should be sufficient, if this is for `sft`

arctic_platform/client/config.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Literal
+
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
+
+JobId = int | str
+
+
+class ArcticRLClientConfig(BaseModel):
```
sfc-gh-truwase, line 29: This should be a superset of current [config](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/main/arctic_platform/rl/config.py), but is currently missing `log_probs_config`, `ds_worker_config`, `arctic_inference_config`, etc.

sfc-gh-truwase (approved):

arctic_platform/client/transports/onprem.py
sfc-gh-mwyatt: Intended for future work

sfc-gh-mwyatt:
> @sfc-gh-mwyatt, I think the `gsm8k` recipe of either [verl](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/main/recipes/rl/verl/simple) or [skyrl](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/main/recipes/rl/skyrl/simple_gsm8k) would be good gating tests for these PRs.

Let's add that in a fast followup after getting #45, #46, and #47 merged.

arctic_platform/client/config.py
sfc-gh-mwyatt: This will need deeper discussion to fully resolve, as we need to figure out how to make neutrino backend + onprem work with nearly identical configs

arctic_platform/client/examples/sft_example.py
sfc-gh-mwyatt: Let's keep it at 8. Reasons will become apparent as I add the cortex backend in #47

sfc-gh-mwyatt:
> main has moved past #36 (the verl integration adapter); all three PRs in the stack are behind. This branch also carries stray commit e697039 (recipes/rl/skyrl … (#32)) which isn't in main and inflates the client-only diff (~866 LOC advertised, +4468 LOC actually on the branch). Rebase before merge, or land #32 first.

@sfc-gh-kganesan I think this is your agent doing something wrong because this is not the case. Ignored, but please confirm manually on your side if you think I'm wrong.

sfc-gh-mwyatt:
> OnPremTransport.shutdown() iterates JOB_TYPES in insertion order rather than reverse creation order. Usually you want to destroy training before its samplers.

@sfc-gh-kganesan Is there a reason to do reverse order? Has this been a problem with onprem deployments previously?

PR #46, opened 2026-07-27
feat(client): add on-prem HTTP transport

sfc-gh-mwyatt:
Add HttpTransport (onprem/dss) alongside the existing Ray transport: it POSTs each op to /{op}?job_id=..., sending an octet torch payload for tensor-bearing ops and JSON otherwise, and can launch/await a local server. Widen the config (comm_protocol http|ray defaulting to http; add the dss backend + HTTP fields) and route make_transport to HttpTransport for onprem-http. Extend the client op tests with make_transport selection for http/dss.

### DIFF
--- a/arctic_platform/client/UNIFICATION_NOTES.md
+++ b/arctic_platform/client/UNIFICATION_NOTES.md
@@ -25,6 +25,15 @@ is built lazily after jobs are initialized. `_rpc` resolves `op -> method` and
 forwards `(job_id, body)` unchanged, matching the server's uniform
 `op(job_id, body) -> dict` surface.
 
+## On-prem HTTP transport
+`HttpTransport` shares the same `OnPremTransport` control plane and only supplies
+the delivery primitives: it POSTs each op to `/{op}?job_id=...`. Tensor-bearing
+ops (`fwd_bwd`/`fwd_no_grad`) send an octet `torch.save` payload and decode octet
+responses; everything else is JSON. It also optionally launches a local server
+(`launch_local_server`) and polls `/health` + `/job/{id}` to wait for readiness.
+The Ray path forwards `(job_id, body)` to the same uniform server surface, so the
+two transports differ only in `_start`/`_rpc`/`_destroy`.
+
 ## Not yet in unified client (future work)
 Ops present in `arctic_platform/rl/*_client.py` but not yet on `ArcticRLClient`.
 Not on the immediate critical path, but they **block deleting the async
--- a/arctic_platform/client/client.py
+++ b/arctic_platform/client/client.py
@@ -30,9 +30,12 @@ from arctic_platform.client.transport import Transport
 
 
 def make_transport(config: ArcticRLClientConfig) -> Transport:
+    from arctic_platform.client.transports.onprem_http import HttpTransport
     from arctic_platform.client.transports.onprem_ray import RayTransport
 
-    return RayTransport(config)
+    if config.backend == "onprem" and config.comm_protocol == "ray":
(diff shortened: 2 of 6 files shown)


sfc-gh-kganesan:
Reviewed the whole three-PR stack end-to-end; full write-up on #47. Notes specific to this PR below.

Ran the tests on this branch: **14/14 pass**.

Small, cohesive addition that reuses #45's `OnPremTransport` base cleanly. Two items to address before merge.

## F2 — `_loads` enables pickle on decode (blocking, one-liner)

`_loads` uses `torch.load(io.BytesIO(data), map_location="cpu", weights_only=False)` at `arctic_platform/client/transports/onprem_http.py:44`. `weights_only=False` re-enables `pickle` on the wire, so a compromised, misconfigured, or wire-buggy server can execute arbitrary code in the client. #47 ships `wire.loads()` with an explicit pickle/zip sniff-and-refuse — please either switch to it, or at minimum set `weights_only=True`. Not gated on the on-prem server change (that's the P2 follow-up on #47); this one is a local safety fix.

## PR description overstates the delivered surface

Description advertises *"on-prem HTTP transport **+ dss backend**"*, but `config.py` here only widens `comm_protocol: Literal["http", "ray"]` — `backend` stays `Literal["onprem"]`. The DSS path actually lands in #47 as `backend="cortex"`. Please update the title/description before merging (or add the `dss` alias here if that's still the intent).

## Plan-level context (P2 on #47)

Reproduced live, byte level:

```
on-prem HTTP frame:  1641 bytes  first-4-bytes: b'PK\x03\x04'         (torch.save zip)
cortex   wire frame:  264 bytes  first-4-bytes: b'\xd0\x00\x00\x00'    (safetensors header)

[wire  -> HTTP loads] FAILS: UnpicklingError: invalid load key, '\xd0'.
[HTTP -> wire  loads] FAILS: WireError: refusing to deserialize a legacy pickle/torch.save payload
```

The two transports speak byte-incompatible wire formats today because the on-prem server still emits `torch.save`. `UNIFICATION_NOTES.md` is honest about the ordering constraint. Follow-up sequence on the plan track: flip `arctic_platform/rl/http_server.py` to DSSST1, then delete `_dumps`/`_loads` from `onprem_http.py`. Until that lands, F2 above at least removes the RCE risk on the client side.

## Not blocking

- `_destroy` swallows all exceptions with a bare `except Exception: pass` (line 83-84). If cleanup silently fails, callers won't know their job leaked. Log it.
- `_launch_server` uses `subprocess.Popen` without capturing stdout/stderr. If the server crashes on startup, only "server timed out" bubbles up. Tee stderr somewhere.
- `base_url = f"http://{config.host}:{config.port}"` — no `https`/`scheme` option. Fine for local, not for any remote deployment.
- No retry adapter on `requests.Session` and no per-request `timeout=`. A wedged server will hang the client indefinitely (only `_is_running` has an explicit 5s timeout).
- No codec-level test for `_dumps`/`_loads` roundtripping a real tensor. Trivial to add.

None gate merge.

sfc-gh-mwyatt:
> so a compromised, misconfigured, or wire-buggy server can execute arbitrary code in the client.

@sfc-gh-kganesan this is how the implementation is in the current onprem http server - considering it's onprem, does arbitrary code execution matter? The client will be running in the same environment as the server

sfc-gh-kganesan:
@sfc-gh-mwyatt you are right! , also as we are considering deprecating on-prem http we can ignore that comment

sfc-gh-truwase (approved):

Issue #49, opened 2026-07-30
Cortex-serverless enablement for verl + SkyRL via the unified ArcticRLClient


sfc-gh-kganesan:
Tracking issue for Cortex-serverless enablement of the two merged upstream RL integrations — [NovaSky-AI/SkyRL#1837](https://github.com/NovaSky-AI/SkyRL/pull/1837) (`integrations/arctic_rl/`) and [verl-project/verl#6422](https://github.com/verl-project/verl/pull/6422) (Arctic-specific `RemoteBackend` in `arctic_platform/integrations/verl/adapter.py`) — **without changing either integration's code**.

## Design principle

Both integrations construct their RL client via exactly one call:

```python
from arctic_platform.rl import ArcticRLClientConfig, create_arctic_rl_client
client = create_arctic_rl_client(config, server_state)
```

Both hardcode `backend="local"` in that construction (SkyRL `config.py:570`, verl `adapter.py:559`), so a yaml-level `backend=cortex` is a silent no-op today. The bridge: an env-var override inside `create_arctic_rl_client` that rewrites the incoming config to Cortex when the launcher exports `ARCTIC_RL_BACKEND=cortex`. Zero adapter change.

### 1. Unified client + Cortex transport (Mike, done or in review)

- [x] [#45](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/45): `arctic_platform.client.ArcticRLClient` + `Transport` ABC + on-prem `RayTransport`.
- [x] [#46](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/46): on-prem `HttpTransport`.
- [ ] [#47](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/47): `CortexTransport` + DSSST1 safetensors wire codec. **Open, MERGEABLE, `mergeStateStatus=BLOCKED`.** #50 stacks on this.

### 2. Compat + `arctic_platform.rl` dispatch + env-var override + fake Cortex GS — [#50](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/50) (draft)

Stacked on #47. All changes on the Arctic-Platform side; zero integration touch.

[...]

PR #50, opened 2026-07-30
feat(rl): Cortex-serverless dispatch via arctic_platform.rl (zero integration change)

sfc-gh-kganesan:
Depends on / stacked on top of [#47](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/47). **Do not merge before #47** — base branch is `mwyatt/unified-client-3` so the diff is only the compat + dispatch + env-override delta.

Tracked by [#49](https://github.com/Snowflake-AI-Research/Arctic-Platform/issues/49).

## Why

Both merged upstream integrations —
[NovaSky-AI/SkyRL#1837](https://github.com/NovaSky-AI/SkyRL/pull/1837) (`integrations/arctic_rl/`) and [verl-project/verl#6422](https://github.com/verl-project/verl/pull/6422) (Arctic `RemoteBackend` in `arctic_platform/integrations/verl/adapter.py`) — construct their client via one call:

```python
from arctic_platform.rl import ArcticRLClientConfig, create_arctic_rl_client
client = create_arctic_rl_client(config, server_state)
```

**Both hardcode `backend="local"` in that construction** (SkyRL config.py line 570, verl adapter.py line 559), so setting `backend=cortex` in yaml is a silent no-op today. This PR delivers Cortex serverless enablement without touching either adapter — the launcher exports env vars and `create_arctic_rl_client` rewrites the config before dispatch.

### Compat surface on `arctic_platform/client/*`

Every call site both integrations reach for is now available on the unified client + Cortex transport.

- `arctic_platform/client/config.py` — `extra="ignore"` (was `"forbid"`) so legacy fields pass through. `_apply_legacy_aliases` collapses `backend={local, dss-platform, neutrino}` → `{onprem, cortex}` (with `DeprecationWarning`), aliases `sample_gpus` → `sampling_gpus`, drops `log_prob_engine` / `sampling_engine` / `reference_model` / `job_name` / `experiment_name`. Adds on-prem parity fields (`log_prob_ds_config`, `ds_worker_config`, `arctic_inference_config`, `full_determinism`) threaded to the on-prem transports via `_init_payload`.
- `arctic_platform/client/client.py` — `training_job_id` / `sampling_job_id` / `log_prob_job_id` / `get_server_state()` properties. `fwd_bwd` / `fwd_no_grad` grow `**legacy_kwargs` folded into the transport body (SkyRL: `post_processors=[...]`; verl: `reference_model=…`, `context=…`, `router_replay=…`). `sync_weights(cuda_ipc=False, low_memory=False)`. All colocation-lifecycle ops (`wake_ / sleep_ × {training, inference, log_prob}`, `empty_training_cache`, `weight_norm`, `save_weights`). `_flatten_metrics` bubbles `metrics.grad_norm` / `.loss` to the top level and mirrors `loss` ↔ `avg_loss`.
- `arctic_platform/client/transport.py` — adds nine colocation-lifecycle ops to the canonical `OPS` set.
- `arctic_platform/client/transports/cortex.py` — `_fwd_no_grad` and `_log_probs` handlers; colocation ops registered as `_colo_noop`; `_normalize_train_body` translates the verl-GRPO `{batch, meta, processing}` envelope into Cortex's RPC style; `_shape_train_response` shims loss-only Cortex responses into the on-prem envelope and aliases `model_outputs` → `batch` so `response["batch"]["log_probs"]` works uniformly.
- `arctic_platform/client/transports/onprem_ray.py` — accepts an optional `server_state=` parameter so verl's driver → Ray-worker reconnect flow can reattach across processes.

### `arctic_platform.rl` → Cortex dispatch

- `arctic_platform/rl/config.py` — `backend: Literal["local", "dss-platform"]` widens to include `"cortex"`. Cortex-only fields added: `cortex_host` / `_database` / `_schema` / `_endpoint` / `_pat_env_var` / `_base_url`, `max_seq_len`. `_derive_host_port` short-circuits on cortex, and now also when the caller passed both `host` and `port` explicitly (avoids an eager ray_cluster → tensordict import chain on CPU-only test envs).
- `arctic_platform/rl/client.py` — `create_arctic_rl_client` early-branches when `config.backend == "cortex"` and returns via `_cortex_dispatch.build_cortex_client`. `ArcticRLHTTPClient` / `ArcticRLRayClient` moved to lazy imports inside the on-prem branch so the Cortex code path never triggers the on-prem server chain.
- `arctic_platform/rl/_cortex_dispatch.py` (new): `_to_unified_config` translates the legacy config to unified shape; `_CortexClientShim` wraps `arctic_platform.client.ArcticRLClient` and re-exposes the legacy async surface both integrations reach for.

### Env-var Cortex override — bridges the adapter-hardcode gap

`create_arctic_rl_client` calls `_maybe_override_from_env(config)` before dispatch. When `ARCTIC_RL_BACKEND=cortex` is set, the incoming (hardcoded-local) config gets rewritten to cortex with fields sourced from the env. This is what makes "zero integration change" actually work today, given the adapter hardcodes.

Recognized env vars:

- `ARCTIC_RL_BACKEND=cortex` — toggle; any other value is a no-op.
- `CORTEX_BASE_URL`, `CORTEX_HOST` — target.
- `CORTEX_DATABASE`, `CORTEX_SCHEMA`, `CORTEX_ENDPOINT`, `CORTEX_PAT_ENV_VAR`.
- `CORTEX_MAX_SEQ_LEN` — int; bad values logged and dropped.

Explicit fields on the input config win over env — env is a fallback for adapters whose yaml doesn't yet thread cortex knobs.

### Lazy `arctic_platform/rl/__init__.py`

Eager exports (pydantic only): `ArcticRLClientConfig`, `WeightSyncConfig`. Everything else (`create_arctic_rl_client`, `WeightSyncCoordinator`, all `processors.*` re-exports) loads via PEP 562 `__getattr__` on first attribute access. Result: a Cortex-only driver needs `arctic-platform` + `pydantic` + `requests`. Nothing else. Pinned by `test_cortex_path_never_imports_onprem_transports` + `test_onprem_transports_are_lazy_at_module_level`.

## Tests

**116 pass locally on a driver with only `pydantic` + `requests` + `fastapi` + `uvicorn` + `safetensors` + `torch` installed. No vllm, no ray, no arctic_inference, no tensordict.**

Breakdown:

- `tests/client/test_skyrl_verl_compat.py` — 26 tests pinning every call site SkyRL + verl make against the unified client (using a `_RecordingTransport` shaped like the on-prem server).
- `tests/client/test_client_ops.py` — updated for the expanded `sync_weights` body and the nine new lifecycle ops.
- `tests/client/test_rl_cortex_dispatch.py` — 29 tests total:
  - `TestConfigTranslation` (5), `TestShimAsyncSurface` (10), `TestShimSyncSurface` (3), `TestPropertySurface` (2), `TestFactoryDispatch` (3).
  - `TestEnvOverride` (6, new): env unset is no-op; env flips local → cortex; every recognized env field threaded; explicit config fields win; non-"cortex" toggle values ignored; bad `CORTEX_MAX_SEQ_LEN` warned and dropped.
- `tests/e2e/test_cortex_transport_smoke.py` (new, 18 tests) — full plumbing E2E: driver → unified client → CortexTransport → real HTTP → `fake_cortex_gs` → back:
  - `TestInitialize` — CreateJob + WaitForJob polling completes; sub-jobs captured.
  - `TestCoreOps` (8) — every op the transport dispatches (`fwd-bwd`, `fwd-no-grad`, `step`, `save-checkpoint`, `generate`, `log-probs`, `sync-weights`, `reset-prefix-cache`) round-trips through real DSSST1-chunked octet upload + response decode. `fwd-no-grad` returns a real logprobs tensor at `response["batch"]["logprobs"]` of shape `(batch, seq_len)` after `model_outputs → batch` alias.
  - `TestColocationNoops` (9) — every colocation lifecycle op registered under `_colo_noop` returns `{}` cleanly.

## The fake Cortex GS

`tests/e2e/fake_cortex_gs.py` speaks every REST route the transport actually hits (`CreateJob` / `GetJob` / `forward-backward` / `forward-no-grad` / `generate` / `step` / `save` / `log-probs` / `operation` / `cancel` / `GET requests/{id}`), decodes DSSST1 chunked uploads via `wire.decode_byte_chunks`, and returns shape-plausible canned responses. Fake for training values (random losses, random logprobs) but **plumbing-honest**: same wire codec, same octet-stream framing, same response envelopes. Doubles as an executable reference for what the real Cortex-training endpoint needs to accept.

Runnable standalone: `python -m tests.e2e.fake_cortex_gs --port 8080`. Test suite uses the in-thread `serve_in_background()` helper.

## User-facing UX (what a launcher does to go serverless)

**Zero code change on either integration.** The launcher exports env vars:

```bash
export ARCTIC_RL_BACKEND=cortex
export CORTEX_BASE_URL=https://cortex.snowflakecomputing.com   # or a mock
export CORTEX_DATABASE=my_db
export CORTEX_SCHEMA=rl
export CORTEX_ENDPOINT=cortex-training
export CORTEX_PAT_ENV_VAR=CORTEX_PAT   # PAT itself in CORTEX_PAT
```

Then `run_gsm8k_grpo_arl.sh` (verl) or `run_gsm8k_grpo_4gpu.sh` (SkyRL) runs unchanged; both adapters still build a `backend="local"` config but the factory rewrites it before dispatch.

## What this validates

- Full client-side chain end-to-end, no mocks at any layer between the driver and the REST endpoint. The 18 smoke tests hit the fake GS via real HTTP over a real socket — no `requests-mock`, no `respx`. Wire codec, chunked upload, response polling, response shim, all exercised.
- Both integrations reach the Cortex path via the env-var override with zero adapter change.
- The on-prem path is byte-identical to `main` and stays lazy; `test_cortex_path_never_imports_onprem_transports` guards the invariant.

## What this does NOT validate

- **Training convergence.** Fake GS returns random losses. Real Cortex server is still the gate for loss-curve validation.
- **Full verl / SkyRL driver runs.** Those need `verl` / `skyrl` + `arctic-inference[vllm]` + `tensordict` installed on the driver box. See #49 §3 for the runtime pieces.

## Follow-ups (see [#49](https://github.com/Snowflake-AI-Research/Arctic-Platform/issues/49))

Server-side (Neutrino GS), blocking real Cortex training convergence:

- `POST /{job_id}/forward-no-grad` accepting `{args, kwargs, meta, processing, reference_model?}` (DSSST1 chunks in) → `{model_outputs: {logprobs, entropy?}, metrics: {...}}` (DSSST1 out).
- `POST /{job_id}/log-probs` (JSON in / DSSST1 out).
- Grow `metrics: {grad_norm, ppo_kl, pg_clipfrac_lower, pg_loss, kl_loss, kl_coef}` block on `fwd_bwd` / `step` responses.
- Batch-contract validation for pre-tokenized `kwargs={input_ids, attention_mask, position_ids, prompts, responses, response_mask, advantages, old_log_probs, ref_log_prob?}` on a real cluster.

The `fake_cortex_gs.py` shipped in this PR speaks the exact interface the real server needs to match — GS team can run their implementation against `test_cortex_transport_smoke.py` for a red/green.

## Test plan

- [x] `pytest tests/client/ tests/e2e/` locally — 116 pass on a Cortex-shaped dev env (no ray/vllm/arctic_inference/tensordict).
- [x] AST/import sanity: `arctic_platform.rl.client` no longer exposes `ArcticRLHTTPClient` / `ArcticRLRayClient` at module scope; the cortex dispatch path never imports the on-prem transport modules.
- [x] Env-var override: full 6-test matrix of env states asserts the rewrite behaves like a normal config path.
- [x] End-to-end plumbing: 18 smoke tests hit the fake GS through real HTTP, real DSSST1 chunked upload, real request-status polling. Every op the transport calls round-trips.
- [ ] On-prem regression smoke against SkyRL's `run_qwen3_0.6b_gsm8k_grpo_arl.sh` — loss curves match `main` to within numerical noise. Runs unchanged with `backend: local` (default). Requires the on-prem stack installed.
- [ ] Cortex first-run against the same recipe with `ARCTIC_RL_BACKEND=cortex` + a real Cortex endpoint, once the server-side items in #49 §3 land.

## Superseded

- ~[#51](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/51)~ — closed. The verl adapter swap it proposed is no longer needed; the env-override + dispatch shim in this PR handles both integrations from the Arctic-Platform side.

### DIFF
--- a/arctic_platform/client/client.py
+++ b/arctic_platform/client/client.py
@@ -42,14 +42,71 @@ def make_transport(config: ArcticRLClientConfig) -> Transport:
     return HttpTransport(config)  # onprem (HTTP)
 
 
+def _flatten_metrics(result: Any) -> Any:
+    """Bring ``result["metrics"]`` keys to the top level for legacy readers.
+
+    SkyRL reads ``client.step().get("grad_norm")`` and verl reads
+    ``client.fwd_bwd()["avg_loss"]`` at the top level, but the on-prem server
+    nests scalars under ``metrics`` and Cortex returns loss-only. Both callers
+    predate the unified schema. Flattening once here means the transports can
+    keep returning their native shape and callers see a superset dict:
+
+        {"loss": ..., "avg_loss": ..., "grad_norm": ..., "metrics": {...}, ...}
+
+    Existing top-level keys always win over ``metrics`` keys of the same name.
+    """
+    if not isinstance(result, dict):
+        return result
+    metrics = result.get("metrics")
+    if not isinstance(metrics, dict):
+        return result
+    merged = {**metrics, **result}
+    if "avg_loss" not in merged and "loss" in merged:
+        merged["avg_loss"] = merged["loss"]
+    return merged
+
+
 class ArcticRLClient:
     def __init__(self, config: ArcticRLClientConfig) -> None:
         self.config = config
         self.transport = make_transport(config)
         self.jobs = self.transport.initialize()
 
+    # ── legacy job-id attributes ─────────────────────────────────────────
+    # SkyRL's entrypoint reads pre_client.training_job_id / sampling_job_id /
+    # log_prob_job_id after initialize(). Preserve that surface as attributes
+    # backed by JobHandles rather than forcing the integration to reach into
+    # `client.jobs.*`.
+    @property
+    def training_job_id(self) -> Any:
+        return self.jobs.training
+
+    @property
+    def sampling_job_id(self) -> Any:
+        return self.jobs.sampling
+
+    @property
+    def log_prob_job_id(self) -> Any:
+        return self.jobs.log_prob
+
+    def get_server_state(self) -> Any:
+        """Expose the transport's server state to callers that reconnect Ray.
+
+        The SkyRL entrypoint reads this after creating the driver-side client
+        and forwards it to the Ray worker so the worker can reattach to the
+        same on-prem Ray server. Non-Ray transports return ``None``.
+        """
+        getter = getattr(self.transport, "get_server_state", None)
+        return getter() if callable(getter) else None
+
     # ── training ─────────────────────────────────────────────────────────
-    def fwd_bwd(self, batch: dict, processing: dict | None = None, router_replay: Any = None) -> dict:
+    def fwd_bwd(
+        self,
+        batch: dict,
+        processing: dict | None = None,
+        router_replay: Any = None,
+        **legacy_kwargs: Any,
+    ) -> dict:
         # NOTE: the call *signature* is unified across backends, but `batch`'s
         # *content* is not: Cortex expects an RPC-style
         # {"args": [...], "kwargs": {...}} that the server tokenizes, while on-prem
@@ -60,12 +117,25 @@ class ArcticRLClient:
         # contract (ideally Cortex's) so this frontend is truly backend-agnostic
         # and callers stop branching on backend. Until then `processing` is folded
         # into the body here, which is why callers can leave it inside `batch`.
+        # Legacy callers (verl `compute_log_prob`, SkyRL wire packer) pass
+        # `post_processors=[...]`, `context={...}` or `kwargs={...}` alongside
+        # `batch`; we fold them into the body so the transports can forward one
+        # uniform envelope.
         body = dict(batch)
         body.update({k: v for k, v in (("processing", processing), ("router_replay", router_replay)) if v is not None})
-        return self.transport.call(Request("fwd-bwd", self.jobs.require("training"), body, binary=True))
+        body.update({k: v for k, v in legacy_kwargs.items() if v is not None})
+        result = self.transport.call(Request("fwd-bwd", self.jobs.require("training"), body, binary=True))
+        return _flatten_metrics(result)
 
-    def fwd_no_grad(self, batch: dict) -> dict:
-        return self.transport.call(Request("fwd-no-grad", self.jobs.require("training"), dict(batch), binary=True))
+    def fwd_no_grad(self, batch: dict, **legacy_kwargs: Any) -> dict:
+        # SkyRL's arctic_generator / _ArcticDispatch calls this with
+        # `post_processors=["logprobs"]`; verl calls it with `context=...` and
+        # `processing=...`. Fold every extra kwarg into the body so transports
+        # receive one uniform envelope (see `CortexTransport._fwd_no_grad`).
+        body = dict(batch)
+        body.update({k: v for k, v in legacy_kwargs.items() if v is not None})
+        result = self.transport.call(Request("fwd-no-grad", self.jobs.require("training"), body, binary=True))
+        return _flatten_metrics(result)
 
     def step(self, learning_rate: float | None = None) -> dict:
         # NOTE: response *shapes* also diverge across backends -- on-prem returns
@@ -74,7 +144,8 @@ class ArcticRLClient:
         # via graceful key lookups today.
         # TODO(unify-backends): converge the server-side fwd_bwd/step *responses*
         # on ONE schema (loss + metrics) so callers never key on backend.
-        return self.transport.call(Request("step", self.jobs.require("training"), {"learning_rate": learning_rate}))
+        result = self.transport.call(Request("step", self.jobs.require("training"), {"learning_rate": learning_rate}))
+        return _flatten_metrics(result)
 
     def save_checkpoint(self, stage_info: dict | None = None, path: str | None = None) -> dict:
         # `path` is this call's destination; when None the server uses the job's
@@ -83,7 +154,12 @@ class ArcticRLClient:
         return self.transport.call(Request("save-checkpoint", self.jobs.require("training"), body))
 
     # ── sampling / log-prob ──────────────────────────────────────────────
-    def generate(self, prompts: list, sampling_params: dict | None = None, routing_key: Any = None) -> list:
+    def generate(
+        self,
+        prompts: list,
+        sampling_params: dict | None = None,
+        routing_key: Any = None,
+    ) -> list:
         body = {"prompts": prompts, "sampling_params": sampling_params, "routing_key": routing_key}
         return self.transport.call(Request("generate", self.jobs.require("sampling"), body))["results"]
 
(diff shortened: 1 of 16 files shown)


sfc-gh-kganesan:
### End-to-end test on a Cortex QA endpoint (CPU-only driver)

Ran the `recipes/rl/skyrl/simple_gsm8k` recipe against a live Cortex QA sub-account with:

- `ARCTIC_RL_BACKEND=cortex` (PAT auth via `CORTEX_HOST` + `CORTEX_PAT_ENV_VAR`)
- Driver on a CPU-only pod (all GPU work delegated to Cortex, as this PR intends)
- Qwen3-0.6B GRPO, KL disabled (`use_kl_loss=false`, `use_kl_in_reward=false`), `n_samples_per_prompt=4`, `train_batch_size=32`, `update_epochs_per_batch=1`

#### What works out of the box
Job spin-up, `save_weights_for_sampler` → `sync_weights` → eval (`generate` + reward scoring) all complete cleanly; wire framing, DSSST1 chunking, and operation polling behave as designed. The env-var-driven backend override (`ARCTIC_RL_BACKEND=cortex`) does redirect through `arctic_platform.rl` without touching SkyRL/Verl adapters — the "zero integration change" claim holds on the client side.

#### Two CPU-only-driver frictions worth guarding
1. `arctic_platform.rl.utils.server_models` eagerly imports `arctic_inference.server.config.ModelConfig`. On a Cortex-only driver `arctic_inference` isn't (and shouldn't need to be) installed — this makes the whole `arctic_platform.rl` import chain fail. Suggest a lazy / `TYPE_CHECKING` import so the utility modules load without the server package on Cortex drivers.
2. `skyrl/train/utils/utils.py::peer_access_supported` unconditionally spins up a Ray placement group requesting `{'CPU':1,'GPU':2}` when `torch.cuda.is_available()` is False, which hangs the autoscaler forever on a CPU-only head. Not this PR's code, but the Cortex path is the first user to actually hit it — worth an `SKYRL_SKIP_P2P_CHECK=1` escape hatch (either in the arctic_rl entrypoint or upstream SkyRL).

#### Convergence blocker (already called out in the "Follow-ups" section)
Training gets through eval + step-0 sync, then dies on step 1 inside `_ArcticDispatch._compute_old_log_probs` → `client.fwd_no_grad` → HTTP 404 because the Cortex GS doesn't yet expose the `forward-no-grad` endpoint. This matches the follow-up list in the PR description exactly — client side is doing its job, purely waiting on server-side op registration.

Worth noting for reviewers: this call is **not** a KL-loss thing (KL is fully off in the recipe, no reference model instantiated). It's the π_old snapshot for the PPO/GRPO ratio, invoked once per training step regardless of KL config.

Once the endpoint lands (or old-log-probs are folded into `forward-backward`'s response metrics), convergence validation can proceed.

PR #53, opened 2026-07-31
Add Model Factory Base

sfc-gh-mwyatt:
First cut of `arctic_platform.model`. Right now every DeepSpeed worker builds its model by hand (`from_pretrained` plus a pile of patches), and we want one place that does it from a model spec. This PR is just the scaffolding so we can start moving that logic over.

The factory builds and configures the model and returns it. It doesn't call deepspeed.initialize, build optimizers, set up process groups, or checkpoint; the worker keeps doing all of that.

How it fits together:
- A `ModelSpec` (pydantic, JSON-serializable) says what to build.
- Loaders are builders registered by name. One is the default, and a loader can
  carry a `matches` predicate so it gets picked automatically (e.g., based on model architecture).
- Optimizations are small functions registered by name and run in one shared
  order list, since they may step on each other. Nothing runs unless you turn it on.
- `build_model(spec)` ties it together: pick the loader, build, run the enabled
  optimizations, return a `LoadedModel`.

Adding the most basic loader (huggingface) and optimization (liger) in this PR to demonstrate what each looks like. Also added EP/SP configs, but they are currently ignored.


arctic_platform/model/optimization.py
```diff
@@ -0,0 +1,57 @@
...
+
+from typing import Any
+from typing import Callable
+
+from arctic_platform.model.loader import LoadedModel
+from arctic_platform.model.loader import LoaderContext
+
+# An optimization mutates the model in place given (model, config, ctx), where
+# config is the matching field on ModelSpec.optimizations.
+Optimization = Callable[[Any, Any, LoaderContext], None]
```
sfc-gh-truwase: Why not strict typing?

```suggestion
Optimization = Callable[[LoadedModel, dict (or Config), LoaderContext], None]
```

arctic_platform/model/optimization.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Canonical apply order; they interact, so this is the single source of truth for
+# sequencing (e.g. fp32_lm_head before chunked_lm_head, compile last). Every
+# registered optimization must appear here.
+OPTIMIZATION_ORDER: tuple[str, ...] = ("liger",)
```
sfc-gh-truwase: Is there a standard method for figuring out how `zorro` or new optimizations fit into the ordering?

arctic_platform/model/patches/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Snowflake Inc.
```
sfc-gh-truwase, line 1: I am curious if `optimizations` -> `patches` is more appropriate long-term. Although, the current patches are optimizations, I am curious if that could ever change.

sfc-gh-truwase (approved):

arctic_platform/model/optimization.py
sfc-gh-mwyatt: Unfortunately I don't think there will be - this is the part I'm most unsure about because it's hard to guarantee compatibility of any combination of optimizations.

arctic_platform/model/patches/__init__.py
sfc-gh-mwyatt: "patches" is a more generic term - I like this!

arctic_platform/model/optimization.py
sfc-gh-mwyatt: Let me fix this!

PR #52, opened 2026-07-30
Align on-prem APIs with SnowAPI payload shapes

sfc-gh-mwyatt:
Refactors the on-prem HTTP/Ray path so the unified `ArcticRLClient` uses one op vocabulary and request-body shape across both the on-prem and (upcoming) Cortex transports. No Cortex transport is added here.

- **Shared DSSST1 binary wire.** New `arctic_platform.wire` (safetensors-based) replaces `torch.save`/`torch.load` across the client, server, `http_client`, and `batch` utils — one codec for all tensor-bearing payloads, and no more pickle in the deserialization path.
- **Unified `/operation` envelope.** Control-plane ops (`weight-sync`, `reset-prefix-cache`) now flow through a single `operation` op carrying Cortex's `{operation_type, sub_job_id, sub_job_type, payload}` shape, with matching `/operation` endpoints on the HTTP and Ray servers.
- **Body shapes match Cortex:** `weight-sync` (`source_sub_job_id`/`target_sub_job_ids`), `save` (`checkpoint_id`/`checkpoint_type`), `step` (`learning_rate`), and `generate` (`prompts`/`sampling_params`/`routing_key`/`strict`).
- **`generate` over octet DSSST1 (HTTP).** The HTTP `/generate` route now speaks the binary wire in/out like `forward-backward`, matching SnowAPI (Ray is in-process, unchanged).
- **Transport cleanup:** collapsed the `_rpc` delivery primitive into `call`.
- **Docs:** `UNIFICATION_NOTES.md` records the alignment plus the divergences we're intentionally deferring (async submit/poll model, job-creation shape, addressing/id types) — handled by the transport layer since `dss-client` will be deprecated in favor of the Cortex transport on the same wire.

PR #54, opened 2026-07-31
Feature: Cortex Transport

sfc-gh-mwyatt:
Follow-up to #52 adding cortex transport.


sfc-gh-kganesan:
## E2E against Cortex QA6

SkyRL GSM8K (GRPO, Qwen3-0.6B, `train_batch_size=8`, `n_samples_per_prompt=4`, `attn_implementation=sdpa`), CPU-only driver. This branch wired into SkyRL `integrations/arctic_rl` via a `_cortex_dispatch` shim that bridges the legacy async surface onto the unified `ArcticRLClient(CortexTransport)`.

Sits on top of #52 (DSSST1 wire, `/operation` envelope, `Transport`/`JobHandles`/`Request` ABC); scope of what the E2E exercised there is reported in the [#52 coverage note](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/52#issuecomment-5160155320).

### Tested E2E

- `POST /cortex-training` with `sub_job_configs` (`training` + `sampling`), `_wait_running`, sub-job routing via `JobHandles`
- `generate` with `logprobs=1` — per-token logprobs used as π_old, avoiding the missing `forward-no-grad` (matches Prime-RL's `LossInputs.inference_logprobs`)
- `fwd_bwd` with the canonical GRPO payload:
  - `kwargs = {input_ids, attention_mask, position_ids}`
  - `context = {input_ids, old_log_probs_shifted, advantages, loss_mask}`
  - `processing = {post: ["compute_logprobs"], loss_fn: "grpo", config: {eps_clip, dp_size, batch_num_tokens, global_batch_size, prox_logp_method}}`
  - Loss runs on ArcticTraining-DSS main.
- `step` returns metrics; step-1: `reward/avg_raw_reward=0.25`, `policy/grad_norm=0.90`, `policy/importance_weight≈1.0` (ratio starts at 1 because rollout logprobs == π_old)
- Next-step `generate`

### Blocking multi-step convergence

`weight-sync` fails at `BroadcastNCCLEngine.__init__` when training and sampling sub-jobs land on different Ray nodes. Reproduced from the SkyRL loop **and** from a bare `dss-neutrino weight-sync` on the same job:

```
File "vllm/distributed/device_communicators/pynccl.py", line 144, in __init__
    self.all_reduce(data)
RuntimeError: NCCL error: internal error - please report this issue to the NCCL developers
```

Server-side fabric issue; nothing in this branch or the shim can address it. Independent of the shape fix below (retested with both env workarounds off).

### Needs fixing on this branch

**1. `CortexTransport._inference_sub_job` flattens vLLM args onto `inference_config` instead of nesting under `vllm_config`.**

`InferenceConfig` has `extra="allow"` so the flat keys are echoed back on `GET`, but `_prepare_sampling_engine_kwargs` reads `inference_config.vllm_config.copy()` — so `gpu_memory_utilization`, `enforce_eager`, `enable_chunked_prefill`, `enable_prefix_caching`, `max_num_seqs` never reach vLLM. Sampling falls back to defaults (GMU ~99%). Confirmed via `WARNING dss.ray_dss.models: InferenceConfig received unknown field(s): [...]` on QA6 sampling zones.

```diff
 def _inference_sub_job(self, job_type: str, n_gpus: int) -> dict:
-    inference = {"max_seq_len": self.config.max_seq_len, "n_gpus": n_gpus}
-    for k, v in (self.config.vllm_config or {}).items():
-        inference.setdefault(k, v)
+    vllm_cfg = dict(self.config.vllm_config or {})
+    vllm_cfg.pop("max_model_len", None)         # dup of inference_config.max_seq_len
+    vllm_cfg.pop("tensor_parallel_size", None)  # dup of inference_config.n_gpus
+    inference: dict[str, Any] = {"max_seq_len": self.config.max_seq_len, "n_gpus": n_gpus}
+    if vllm_cfg:
+        inference["vllm_config"] = vllm_cfg
     return self._sub_job(job_type, {"inference_config": inference})
```

Post-fix, QA6 accepts the nested shape and drops the `unknown field(s)` warning.

**2. `arctic_platform.rl` eagerly imports `arctic_inference.server.*`, which breaks CPU-only drivers on the Cortex path.**

Lazy-import the HTTP/Ray backends inside their `if config.backend == ...` branches. Without this, `arctic_platform.rl` fails to import on a Cortex-only driver where `arctic_inference` isn't installed.

arctic_platform/client/config.py
```diff
@@ -62,6 +62,18 @@ class ArcticRLClientConfig(BaseModel):
         ),
     )
 
+    # cortex (SnowAPI). Provide cortex_base_url for a direct/mock URL (no auth), or
+    # cortex_host + a PAT in the env var for Snowflake programmatic-access auth.
+    cortex_base_url: str | None = Field(None, description="cortex: direct/mock GS URL; bypasses PAT auth.")
```
sfc-gh-truwase: Is it possible to nest the configs such as below

```python
class OnPremClientConfig(...): 
     ...

class CortexClientConfig(...):
   ...

class ArcticRLClientConfig(...):
    ...
    backend_config: Union[OnPremClientConfig|CortexClientConfig]
``` 

I think if possible, this could improve maintenance and readability

sfc-gh-mwyatt: Great idea, I've updated the implementation!

sfc-gh-truwase (approved):

sfc-gh-kganesan:
Re-reviewed after your rebase onto #58/#59/#62/#63.

**Both blockers from my [Aug 2 comment](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/54#issuecomment-5160155320) are fixed.** `_cortex_inference_sub_job` now pops `max_model_len`/`tensor_parallel_size` and nests the remainder under `vllm_config`, so `gpu_memory_utilization`, `enforce_eager`, `enable_chunked_prefill`, etc. actually reach vLLM — the silent-fallback case is gone.

The sync/async parallelism (`_send`/`_asend` sharing one retry policy over two exception hierarchies), the aiohttp session-per-loop rebuild, and the `_is_connect_error`-only retry on job creates all look right to me. `404`/`409` in the transient set with the GS/ZMD-restart comment is worth keeping.

**One thing worth fixing before merge:** missing PAT env var raises a bare `KeyError` from inside `_auth_headers`. Wrap it:

```python
try:
    token = os.environ[cx.pat_env_var]
except KeyError as exc:
    raise RuntimeError(
        f"Cortex PAT env var {cx.pat_env_var!r} is not set; "
        f"export it or configure cx.base_url for a local/dev host."
    ) from exc
```

This is the first error users hit when `CORTEX_PAT` is misspelled, and today it doesn't tell them what to do.

**Shim-reduction stack I put on #55** — cherry-pick or ignore, both contained to `CortexTransport.call`/`acall`, no frontend change, no on-prem impact:

- [`548b14c`](https://github.com/Snowflake-AI-Research/Arctic-Platform/commit/548b14c) — lift the training-scalar reshape (`avg_loss → metrics.loss` + a small set of other scalars) into a new `_normalize_response` in the transport, so every `arctic_platform.client` user gets the on-prem-shaped response (not just SkyRL/verl via the shim). Consequence: [`ad948b0`](https://github.com/Snowflake-AI-Research/Arctic-Platform/commit/ad948b0) drops `_normalize_step` from `_cortex_dispatch` (−22 LOC).
- **Not done, proposal:** the fwd_bwd *request*-shape reshape (SkyRL/verl `{batch, meta, processing}` → Cortex `{args, kwargs, context, processing}`) could move into `CortexTransport` too. The frontend `_fwd_bwd_request` docstring's `TODO(unify-backends)` is exactly this. Would drop another ~50 LOC from the shim. I held off because it might belong server-side (accept both shapes) — happy to send whichever you'd rather.

**What I actually tested:** CPU smoke test only — fake `Transport` driving `_CortexClientShim` and `_normalize_response`, verifying fwd_bwd payload shape, step/fwd_bwd response reshape, sync-shutdown path. No QA6 re-run on this rebase; my earlier QA6 numbers were pre-rebase, before #58/#59/#62/#63.

The BroadcastNCCL init failure I hit on the previous QA6 run (training and sampling sub-jobs on different Ray nodes) was a fabric issue, unrelated to this PR — flagging so it doesn't get attributed if it recurs.

PR #55 is rebased on top of your current head; once #54 merges I'll rebase onto main and drop the `origin/mwyatt/*` dependency.

sfc-gh-mwyatt:
> 1. PAT missing env var raises KeyError — surface a clearer message.

This is already present in the config validation:
```python
        if not (self.base_url or self.host):
            raise ValueError("cortex: set base_url (direct URL) or host (PAT auth).")
        if self.host and not self.base_url:
            if not (self.database and self.schema_):
                raise ValueError("cortex: database + schema required for host/PAT auth.")
            if not self.resolve_pat():
                raise ValueError(f"cortex: no PAT — set `pat` or the '{self.pat_env_var}' env var for host auth.")
        return self
```

> 2. Consider handling torch import errors in _to_python.

Torch will be a requirement

PR #55, opened 2026-08-04
feat(cortex): SkyRL on Cortex, rebased onto the unified client

sfc-gh-kganesan:
Makes SkyRL RL recipes runnable against Cortex-training GPUs with SkyRL unmodified. The driver stays CPU-only; training fwd/bwd/step, sampling, and weight-sync are dispatched to Cortex sub-jobs through the unified client from #93.

Scope is SkyRL only. An earlier revision of this branch also carried a verl adapter; it was dropped rather than shipped unvalidated, so nothing here changes verl.

## What's in it

- **SkyRL entrypoint** (`integrations/skyrl/entrypoint.py`) — the recipe names it as `trainer.override_entrypoint`, and it swaps upstream's client factory for one built on `arctic_platform.client`. Naming the entrypoint is what selects Cortex; no env var decides it, and `arctic_platform/rl` is untouched.
- **Client shim** (`integrations/_cortex_dispatch.py`) — presents the client surface SkyRL expects over `AsyncArcticRLClient`.
- **Payload lowering** (`integrations/_cortex_shared.py`) — reshapes SkyRL's batch to Cortex's wire format and left-aligns every row, which the server's microbatch packer requires. This was the substantive correctness fix; see below.
- **Recipe** (`recipes/rl/skyrl/simple_gsm8k_cortex/`) — GSM8K GRPO on Qwen3-0.6B at the on-prem sibling's hyperparameters, unchanged, with preflights for the failure modes that otherwise surface as opaque 429s.
- **Config** — `CortexConfig` becomes a `pydantic-settings` model reading `ARCTIC_CORTEX_*`.
- **Transport** — `list_jobs` / `cancel_job`, plus a fix where `:cancel` returning an empty body made a successful release report `could not release <id>`.

## Validation

One epoch of GSM8K at the shipped defaults, all 233 steps, no overrides and no early stop — a bare `./run_qwen3_0.6b_gsm8k_grpo_cortex.sh`. Held-out `eval/all/pass_at_1` over the full 1319-example test set:

| step | 0 | 20 | 40 | 60 | 80 | 100 | 130 | 160 | 190 | 210 | 230 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| pass@1 | 0.2942 | 0.4996 | 0.6687 | 0.6839 | 0.7013 | 0.7180 | 0.7445 | 0.7536 | 0.7627 | 0.7566 | 0.7680 |

[...]

PR #57, opened 2026-08-05
Port verl adapter to unified ArcticRLClient

sfc-gh-truwase:
## Summary
- Port `integrations/verl/adapter.py` onto `arctic_platform.client.ArcticRLClient` (async→sync via `run_in_executor`), with Ray reconnect/`get_server_state` parity for forwarder and rollout workers.
- Extend the unified on-prem client/config/RayTransport for the ops the adapter needs: init extras (`ds_worker_config`, `arctic_inference_config`, …), `fwd_no_grad(reference_model=…)`, `sync_weights(cuda_ipc, low_memory)`, `sleep_inference`/`wake_inference` via `/operation`, dedicated-thread RayTransport loop + pickling for actor handoff.
- Round DeepSpeed `warmup_num_steps` to an integer so ZeRO schedulers accept `warmup_ratio * horizon`.

## Test plan
- [x] 1-GPU GSM8K smoke (`total_training_steps=1`) via autorun: unified client path, cuda_ipc weight-sync, sleep/wake, `step:1`, EXIT 0
- [x] 8-GPU full GSM8K run (`job-20260805T181456Z`, Qwen3-1.7B, 15 epochs / 3495 steps, colocated train+sample): still in progress at ~33% (step ~1169); clear convergence — train reward ~0 → ~0.93, val `gsm8k/acc` ~0.4% → **~88–90%** (e.g. step 1130 ≈ 89.8%)
- [x] Reconnect / Ray actor handoff: `ArcticLLMServer` receives pickled reconnect handle; dedicated-thread RayTransport `__getstate__`/`__setstate__` avoids uvloop serialize errors (exercised in smoke + 8-GPU run)


sfc-gh-kganesan (approved):
Reviewed all 9 files, ran the tests locally, and validated the config-construction path against my PR #55 shim. LGTM — safe to merge.

## What each piece does (as a note for reviewers landing on the PR fresh)

* `arctic_platform/client/client.py` — `make_transport(config, server_state=None)`; `_fwd_no_grad_request(..., reference_model=False)` routes to `log_prob` when `reference_model=True`, `training` otherwise (mirrors pre-refactor `ray_client.fwd_no_grad`); both sync + async clients get `get_server_state()` delegating to the transport (raises `NotImplementedError` for transports that don't own one).
* `arctic_platform/client/transports/onprem_ray.py` — three real changes here:
    * `RayTransport(config, server_state=...)` skips `create_arctic_rl_ray_server_state` and reuses the passed-in actor handle (reconnect mode).
    * `__getstate__ / __setstate__` drop the private event loop before cloudpickle (verl hands the reconnected backend to `ArcticLLMServer` Ray actor; a bound loop is unpicklable) and rebuild a fresh one on the receiving side. Ray handles pickle fine.
    * `_run_on_loop(coro)` — if there's already a running loop in the current thread (verl's `SyncArcticRLClient.shutdown` inside `asyncio.run`) hop to a worker thread so we don't nest `run_until_complete`.
* `arctic_platform/client/transports/onprem_http.py` — reconnect clients (`JobHandles.from_config(config).any_set`) skip spawning a local server; forwarders / rollout replicas must dial the existing one.
* `arctic_platform/common/deepspeed_worker.py` — `round(warmup_ratio * horizon)`. DeepSpeed's `WarmupLR/WarmupCosineLR` require integer `warmup_num_steps`; a fractional value silently misconfigures the scheduler.
* `arctic_platform/common/http_server.py` and `arctic_platform/common/ray_server.py` — extend `/operation` dispatch to `sleep-inference` / `wake-inference` (was weight-sync + reset-prefix-cache only). Unifies control-plane ops across both transports.
* `arctic_platform/common/utils/server_models.py` — `OperationRequest.sub_job_id: int | str | None` (int on-prem, str on Cortex).
* `arctic_platform/integrations/verl/adapter.py` — the port itself. Config moves from flat `arctic_platform.rl.ArcticRLClientConfig(backend="local", host=..., training_config={...})` to nested `arctic_platform.client.ArcticRLClientConfig(backend_config=OnPremConfig(...), training=TrainingConfig(ds_config=...), sampling=SamplingConfig(...))`. Optimizer + LR scheduler fold into `ds_config` (canonical shape from #59); `attn_implementation` moves to `ds_worker_config`; HTTP auto-launches the local server.
* `tests/client/test_client_ops.py` — 8 new tests exercising `reference_model` routing, `get_server_state` behavior, and `server_state` threading through `make_transport` / `create_arctic_rl_client`. Fake-transport style is consistent with the existing tests; coverage of the new codepaths is genuine, not perfunctory.

## Validation

* CI green (`unit-tests`, `formatting`, `semgrep`).
* Local `pytest tests/client/` on `448fa37` — 41 passed.
* Constructed both `backend_config=OnPremConfig(...)` and `backend_config=CortexConfig(...)` configs against this branch's nested surface — accepted, `make_transport` routes correctly.
* Existing 8-GPU GSM8K on-prem run cited in the PR description (train reward ~0 → 0.93, val `gsm8k/acc` → 88–90%) is a strong signal the port doesn't regress convergence.

## Non-blocking observations

* **`fwd_no_grad(reference_model=True)` when `log_prob_gpus == 0`** raises inside `jobs.require("log_prob")`. That's the intended behavior (no ref model → can't ask for ref log-probs), but the resulting error message points at the require rather than at the caller's misconfiguration. Not blocking; a one-line guard in `_fwd_no_grad_request` would give a clearer message if desired.
* **Async `sync_weights` docstring** documents "staged wake → operation → wake → reset" — which is right for on-prem colocated, but the same code path currently crashes on Cortex (Cortex has no `wake-inference` op; `CortexTransport._asubmit` hits `body.items()` on a list). My PR #55 shim's `sync_weights` bypasses this by dispatching the raw `operation` request. Fine for now — orthogonal to #57 — but worth teasing apart in a follow-up (either make `sync_weights` composable so backends can opt out of the wake wrap, or make `CortexTransport` accept the `wake-inference` request and no-op).
* **`log_prob_engine`** is now sourced from `config.sampling.log_prob_engine` in `RayTransport.__init__` (was hardcoded `"deepspeed"`). `SamplingConfig.log_prob_engine` defaults to `"vllm"`, so any existing recipe that omitted it would silently switch behavior. verl's adapter port sets it to `"deepspeed"` explicitly (matches old default), so no regression here — but worth flagging for anyone else building configs directly against the unified client.

## Merge recommendation

**Merge as-is.** No blockers. Small nits above are follow-up material, not landing gates.

## What lands on my PR #55 after this merges

* Delete `_CortexClientShim._to_unified_config` (verl no longer feeds legacy `arctic_platform.rl.ArcticRLClientConfig`).
* Delete `_CortexClientShim.fwd_bwd` payload reshape (verl now builds Cortex-canonical `{args, kwargs, context, processing}` against `ArcticRLClient` directly).
* Delete the verl-side of every shim delegator (generate, sync_weights override, save_checkpoint, etc.).
* Keep the SkyRL leg — SkyRL's `integrations/arctic_rl/entrypoint.py` still builds legacy `arctic_platform.rl.ArcticRLClientConfig`, so `ARCTIC_BACKEND=cortex` gate + shim stay until SkyRL gets a similar port.

Happy to open the shim-reduction PR immediately once this lands.

PR #59, opened 2026-08-06
Unified ArcticRLClient better nested configs and validation

sfc-gh-mwyatt:
Restructure the ArcticRLClientConfig to have more nested pydantic models and validate input values. Also add helper methods to massage the configs into the expected shape for the onprem backend. This should eventually go away, but is a convenient helper until we change the server expected configs.

### DIFF
--- a/arctic_platform/client/config.py
+++ b/arctic_platform/client/config.py
@@ -53,7 +53,9 @@ class OnPremConfig(BaseModel):
     port: int = Field(8000, description="onprem: server port.")
     colocate: bool = Field(False, description="onprem: colocate job types on shared GPUs.")
     launch_local_server: bool = Field(False, description="onprem: spawn a local server before connecting.")
-    startup_timeout: float = Field(600.0, description="onprem: seconds to wait for a launched server to become healthy.")
+    startup_timeout: float = Field(
+        600.0, description="onprem: seconds to wait for a launched server to become healthy."
+    )
 
 
 class SamplingConfig(BaseModel):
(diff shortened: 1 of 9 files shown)


arctic_platform/client/examples/sft_example.py
```diff
@@ -66,33 +69,33 @@ def _onprem_config(comm_protocol: str, launch_local_server: bool) -> Callable:
...
-                    "offload_param": {"device": "none"},
+            backend_config=OnPremConfig(
+                comm_protocol=comm_protocol,
+                launch_local_server=launch_local_server,
+            ),
+            training=TrainingConfig(
+                optimizer=OptimizerConfig(lr=LR, weight_decay=0.0, betas=[0.9, 0.999]),
+                lr_scheduler={"warmup_ratio": 0.0},
+                training_horizon=1,
+                gradient_accumulation_steps=1,
```
sfc-gh-truwase: Flagging this duplication of `gradient_accumulation_steps` here and in ds_config.

arctic_platform/client/config.py
```diff
@@ -12,61 +12,213 @@
...
+    )
+    log_prob_engine: Literal["vllm", "deepspeed"] = Field(
+        "vllm", description="Engine backend for the log-prob job (deepspeed builds a forward-only DS engine)."
+    )
+    log_prob_ds_config: dict[str, Any] | None = Field(
+        None, description="Log-prob DeepSpeed worker config, used only when log_prob_engine='deepspeed'."
+    )
+
+
+class OptimizerConfig(BaseModel):
```
sfc-gh-truwase: Why do we need a separate `OptimizerConfig` instead of using (or extending) existing DS [field](https://www.deepspeed.ai/docs/config-json/#optimizer-parameters)? Is there a scenario of constructing the optimizer outside of DS? If so, that probably won't scale since zero-sharding won't be available.

sfc-gh-truwase (approved):

arctic_platform/client/config.py
sfc-gh-mwyatt: Good point - I was thinking it would be nice to validate more config fields with defined pydantic models. But looking at this again, we should just let DeepSpeed handle any validation and rely on the ds config field

PR #61, opened 2026-08-07
Add recipes

sfc-gh-jaelee:
Adding SFT/RL basic recipes, similar to tinker-cookbook.


cortex-client/recipes/sft_loop.py
```diff
@@ -0,0 +1,243 @@
...
+    "down_proj",
+)
+
+
+@chz.chz
+class Config:
+    config: str
+    job_id: str | None = None
+
+    model_name: str = "Qwen/Qwen3.6-35B-A3B"
```
sfc-gh-mwyatt: If the default model is MoE, then we need to change the model_provider and EP defaults too. We should probably just have this be the dense 8B as default.

cortex-client/recipes/README.md
```diff
@@ -0,0 +1,125 @@
...
+| `rl_loop.py` | Math RL (GRPO) | Hendrycks MATH train; MATH-500 held out for eval |
+
+## 1. Prerequisites
+
+```bash
+uv pip install -e cortex-client
+uv pip install 'tinker-cookbook[math-rl] @ git+https://github.com/thinking-machines-lab/tinker-cookbook.git@nightly'
+```
+
+We assume you have the config we've provided under `recipes/config.json`.
```
sfc-gh-mwyatt: Should we add a `recipes/config.json.template`?

cortex-client/recipes/rl_loop.py
```diff
@@ -0,0 +1,579 @@
...
+    # Evals. 0 disables; otherwise baseline at batch 0 and the final batch.
+    eval_every: int = 10
+    # MATH-500 is 500 problems — None means the full split.
+    n_test: int | None = None
+    # uses the training temperature / max_tokens by default.
+    eval_temperature: float | None = None
+    eval_max_tokens: int | None = None
+
+    log_path: str = "/tmp/dss-examples/rl-loop"
+    wandb_project: str = None
```
sfc-gh-mwyatt:
```suggestion
    wandb_project: str | None = None
```

cortex-client/recipes/sft_loop.py
```diff
@@ -0,0 +1,243 @@
...
+    train_on_what: renderers.TrainOnWhat = renderers.TrainOnWhat.ALL_ASSISTANT_MESSAGES
+    pad_to_max_length: bool = False
+    max_steps: int = 100
+
+    # 0 = dense FT. Set e.g. 32 for LoRA (r == alpha).
+    lora_rank: int = 32
+    debug_image_tag: str | None = None
+
+    log_path: str = "/tmp/dss-examples/sft-loop"
+    wandb_project: str = None
```
sfc-gh-mwyatt:
```suggestion
    wandb_project: str | None = None
```

cortex-client/recipes/neutrino_common.py
```diff
@@ -0,0 +1,303 @@
+from __future__ import annotations
+
+import contextlib
+import io
+import json
+import logging
+import os
```
sfc-gh-mwyatt:
```suggestion
```

Unused import

cortex-client/recipes/neutrino_common.py
```diff
@@ -0,0 +1,303 @@
+from __future__ import annotations
+
+import contextlib
+import io
```
sfc-gh-mwyatt:
```suggestion
```

Unused import

cortex-client/recipes/sft_loop.py
```diff
@@ -0,0 +1,243 @@
...
+    n_train_batches = len(train_dataset) // config.batch_size
+    n_dropped = len(train_dataset) % config.batch_size
+    if n_dropped:
+        logger.info(
+            f"Dropping last {n_dropped} examples to keep batch size uniform at "
+            f"{config.batch_size}"
+        )
+    total_steps = (
+        n_train_batches if config.max_steps is None else min(n_train_batches, config.max_steps)
+    )
```
sfc-gh-mwyatt:
```suggestion
    total_steps = min(n_train_batches, config.max_steps)
```

Based on the config, `max_steps` is always a defined int - so no need to check if it's `None`

sfc-gh-mwyatt (changes requested):

cortex-client/recipes/sft_loop.py
sfc-gh-jaelee: Great, MoE, if with LoRA, was working out of the box without model_provider and EP default (but for full FT they needed). But I agree it would be better to use dense 8B model. Will move toward that

cortex-client/recipes/README.md
sfc-gh-jaelee: sgtm, I'll add it

sfc-gh-mwyatt (approved):

PR #64, opened 2026-08-07
Stas/refactor 2

sfc-gh-sbekman:
This is Part 2 of the split off from https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/60

## Summary

Intermediary plumbing PR after `#62` (rl→common relocate). No SFT package yet — that lands in a follow-up.

- Extract tiled logits / CE kernels from `qwen_model_patcher` into `common.utils.tiled_logits` (`none` / `compute` / `memory`); RL imports them back.
- Shared `common.registry` for loss / post-processor registration (used by RL pipeline; SFT will register later).
- Shared `common.utils.weight_sync` helpers; HTTP/Ray servers use them instead of inlined ArcticInference imports where possible.
- `finalize_fwd_bwd_metrics` for token-exact `avg_loss` on fwd-bwd responses (HTTP + Ray).
- `prune_checkpoint_dirs` util (+ unit test) — **not** wired to save/load API yet (checkpoint extras in SFT PR).
- Client: `server_cuda_visible_devices`, sleep/wake inference+training, staged `sync_weights` wake flow.
- HTTP log-prob workers honor `MASTER_PORT` (was hardcoded `29501`).
- Base deps: `ray` / `uvicorn` (servers import them unconditionally); `[rl]` keeps vLLM/sampling stack only.
- Docs: `docs/common.md`, `docs/rl.md`, `docs/index.md` (SFT marked forthcoming). 
- `tests/AGENTS.md` - this is new to help guide agents write consistent tests across the repo
- FA2 once-patcher test skips when real `flash_attn` is unavailable.


arctic_platform/client/transports/onprem_http.py
```diff
@@ -121,45 +124,61 @@ def _wait_running(self) -> None:
...
+        try:
+            self._poll(
+                lambda: self.session.get(f"{self.base_url}/health", timeout=self.timeout).ok,
+                bc.startup_timeout,
+                "server",
+            )
+        except Exception:
+            # Health timeout runs in __init__ (before client init guards); kill the orphan.
+            self._terminate_server()
+            raise
```
sfc-gh-mwyatt, line 178: Should we include an error message here? Or perhaps at the `_launch_server` call-site indicating there was a problem starting the server?

sfc-gh-sbekman: I'm not sure. @sfc-gh-truwase - what do you think?

could also start somewhere and then change if it proves not to be the best place?

sfc-gh-truwase: Error message should be a minimum. Ideally, this should propagate back to client as API call failure

sfc-gh-truwase (approved):

arctic_platform/client/transports/onprem_http.py
sfc-gh-sbekman: thank you for the suggestions, guys - addressed.

PR #65, opened 2026-08-10
RFC: TRL backend for Arctic RL

kashif:
Sketch of a TRL backend for Arctic RL, opened as an RFC alongside huggingface/trl#6676. Not ready to merge: the batch layout helpers are stubs and none of this has run against a live server.

The idea is that TRL keeps the loss. `forward` returns per token logprobs, TRL builds its loss on them, and `backward` sends back `d(loss)/d(logprobs)`. The server backprops `sum(w * logprobs)`, which gives the same parameter gradients as the real loss. Verified bit exact against the in process path on TRL's GRPO loss.

So one server side loss function instead of one per algorithm. For comparison, `integrations/verl/grpo_loss.py` is 523 lines and still covers `vanilla` only, out of the 12 in verl's `POLICY_LOSS_REGISTRY`. `integrations/trl/loss.py` is about 6 lines and would not need touching when TRL adds a loss variant.

Two things I could not settle from the code:

1. `pipeline._resolve_fn` falls back to a dotted path import, so on a self hosted server the loss resolves without a registry entry at all. Is that a supported extension point or an implementation detail?
2. For managed deployments, would a generic registry entry be ok? Weighted cross entropy is the same surrogate under `weights = -d(loss)/d(logprobs)`, which is how Tinker avoids adding a loss.

Remaining open items are in the README. The main one is TRL side: it still wants a local model to prepare and to stream weights from.

### DIFF
--- a/arctic_platform/integrations/trl/README.md
+++ b/arctic_platform/integrations/trl/README.md
@@ -5,33 +5,36 @@ layout helpers in `client.py` are stubs and nothing here has run against a live 
 
 ## The interface
 
-TRL's `TrainingClientProtocol` is two methods:
+TRL's `TrainingClientProtocol` is a single fused call:
 
 ```python
-forward_no_grad(model, input_ids, position_ids, completion_mask, aux_loss_coef) -> ForwardOutput
-forward_backward(grad_log_probs) -> None
+forward_backward(model, input_ids, position_ids, completion_mask, loss_fn, aux_loss_coef) -> ForwardBackwardOutput
 ```
 
-They map one to one onto `fwd_no_grad` and `fwd_bwd`. The server forwards the batch twice, once to score it
-and once to backpropagate, because the graph from the first pass cannot cross the wire.
+`loss_fn` is a Python callable mapping per-token log probs to a scalar. It closes over the advantages, old log
+probs, mask and clipping bounds, so nothing algorithm-shaped reaches this repo.
 
-The trainer computes its loss on the log probs returned by the first call and hands back
-`d(loss)/d(log_probs)`.
-The server backpropagates `sum(w * logprobs)`, a first-order surrogate whose gradient with respect to every
-parameter equals the gradient of the real loss. The loss itself never crosses the wire.
+The adapter runs in the trainer's process even though the model does not, so `loss_fn` is evaluated here, on
+the tensors the server returned. Only tensors cross the wire. The adapter takes `d(loss)/d(log_probs)` locally
+and ships that, and the server backpropagates `sum(w * logprobs)`, a first-order surrogate whose gradient with
+respect to every parameter equals the gradient of the real loss.
 
-Verified bit-exact against the in-process path on TRL's GRPO loss: zero difference across all 27 parameter
-tensors, for both the co-located variant and the two-forward variant a remote backend would use.
+Verified bit-exact against the in-process path on TRL's GRPO loss: zero difference across every parameter
+tensor, for both the co-located variant and the two-forward variant a remote backend would use.
 
-## Why not a fused `forward_backward(..., "grpo_loss")`
+## On the fused call and the extra forward
 
-That was the suggestion on the PR thread, and `forward_backward` here does issue the fused `fwd_bwd`. The split is in
-TRL's interface, not in what the backend does. Two round trips are inherent either way, because the weights
-are a function of the log probs and cannot be known before the first forward. Tinker's
-`forward` + `forward_backward_custom` has the same shape.
+Folding the forward and backward into one call was the request on the PR thread, and this is that call. Naming
+the *loss* in it is the part worth avoiding, and passing a callable instead of a string key gets the fused
+shape without it.
 
-The part worth avoiding is naming the *loss* across the wire. The cost of that is already measurable in this
-repo:
+On cost: a co-located deployment runs one forward, calls `loss_fn` in-process, and returns a loss still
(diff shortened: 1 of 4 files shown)


arctic_platform/integrations/trl/client.py
```diff
@@ -0,0 +1,161 @@
...
+    ) -> ForwardOutput:
+        # `model` is the trainer's local module. Arctic owns the weights, so it is unused here.
+        #
+        # AsyncGRPOTrainer packs a rank's samples into one padding-free row and marks sequence boundaries by
+        # position_ids resets. The server wants padded [B, S] rows, so unpack first. This conversion is the
+        # bulk of the adapter and is the main reason it belongs on this side of the wire.
+        batch = _unpack_to_padded_rows(input_ids, position_ids, completion_mask)
+        self._batch = batch
+
+        response = self.client.fwd_no_grad(
```
sfc-gh-sbekman, line 83: how would bwd work if fwd is no_grad?

I think this `forward` method should contain no_grad -> `forward_no_grad` - otherwise the API is confusing, no?

arctic_platform/integrations/trl/client.py
```diff
@@ -0,0 +1,161 @@
...
+        #
+        # The server returns log probs under the roll(-1) convention and TRL expects the [:, 1:] shift, so
+        # the repack aligns them rather than merely reshaping.
+        return ForwardOutput(
+            log_probs=_repack_to_row(out["logprobs"], completion_mask).requires_grad_(True),
+            entropy=_repack_to_row(out["entropy"], completion_mask),
+            aux_loss=None,  # Arctic reports the MoE aux loss as a metric, not as a differentiable tensor
+        )
+
+    def backward(self, grad_log_probs: torch.Tensor) -> None:
```
sfc-gh-sbekman: same here - shouldn't the API say it's `forward_backward`?

sfc-gh-sbekman (commented):
You're fast, Kashif!

I have some questions:

arctic_platform/integrations/trl/client.py
kashif: Renamed, here and in the TRL protocol, so it is `forward_no_grad` and `forward_backward` on both sides now.

On how bwd works: `fwd_bwd` runs its own forward on the server, so the graph is rebuilt there and the backward is real. The scoring pass cannot carry a graph across the wire, which is the only reason there are two forwards. Added a comment at both call sites since that was not obvious from the code.

arctic_platform/integrations/trl/client.py
kashif: Done, `forward_backward` now.

arctic_platform/integrations/trl/client.py
sfc-gh-sbekman: thank you, Kashif

looks good to me now

Are you planning to merge this already or do more work on this - wasn't sure if to approve it or wait

arctic_platform/integrations/trl/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## What TRL changes
+
+Nothing. `training_client=` already exists on `AsyncGRPOTrainer` alongside the `rollout_worker` and
+`weight_transfer` hooks, and `transformers.Trainer` already accepts `optimizers=`, which is why there is no
+`optim_step` or `clip_grad_norm` in the protocol.
+
+## Open items
+
+1. **No local model.** The real blocker. Arctic owns the weights, but transformers and accelerate still want
```
sfc-gh-truwase, line 71: @kashif, yes this is intentional and a core design principle. The components (trainer, rollout, ..) reside on the server side. For example, the trainer uses [transformers](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/9e748ec3ebdb88f663066540c41b81368b7d4405/arctic_platform/common/deepspeed_worker.py#L163) for model preparation and [liger](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/9e748ec3ebdb88f663066540c41b81368b7d4405/arctic_platform/common/deepspeed_worker.py#L173-L175) for patching. This allows TRL to run on CPU-only machines. Would this work?

arctic_platform/integrations/trl/README.md
```diff
@@ -0,0 +1,83 @@
...
+   a local module to `prepare()`, and `AsyncGRPOTrainer._sync_weight` streams from
+   `self.model.named_parameters()` on every rank (`async_grpo_trainer.py:1149`). Meta device crashes there
+   and an empty CUDA model wastes full model memory. A stub module carrying names and shapes is the likely
+   answer. TRL-side work.
+2. **Microbatch scaling.** The adapter folds the trainer's gradient-accumulation scaling into the weights
```
sfc-gh-truwase, line 76: This is an important point that I believe should be handled by an explicit contract: server-side vs client-side scaling.

sfc-gh-sbekman:
Kashif, one important thing we need to discuss is that in order to do fwd(server) + loss(client) + fwd/bwd(server) approach correctly - we must replay the 2nd forward exactly the same  as the first one - which means, the same RNG engine and if it's an MoE model we also need a router replay for the 2nd fwd.

This is our issue to solve, but I'm just bringing it up that w/o it that Tinker algo won't work correctly

kashif:
Superseded by #84, which carries these three commits rebased and takes them to a validated end to end run, so closing this to keep the review in one place.

@sfc-gh-sbekman that also answers your question above about whether to approve or wait: this one was never going to merge on its own, the batch layout helpers were stubs.

PR #66, opened 2026-08-10
harbor: Cortex training backend as a Harbor plugin (draft)

sfc-gh-kganesan:
Draft. Stacks on top of #55 (Cortex dispatch shim + SkyRL/verl+Cortex recipes) — please merge that first, then re-target this PR to `main`.

## What this PR does

Two things:

1. Adds `arctic_platform/integrations/harbor/` — a Harbor plugin that trains models end-to-end using Harbor's own CLI as the driver, with Cortex Training as the RL backend. Same in-tree subpackage shape as `arctic_platform/integrations/verl/`.
2. Adds an OpenAI-compatible HTTP surface (`/v1/chat/completions`, `/v1/completions`, `/v1/models`) to the sampling sub-job's existing FastAPI server, so any Harbor `BaseAgent` that speaks OpenAI-chat (Terminus 2 via LiteLLM, LangChain-backed agents, custom `BaseAgent`s) can drive rollouts without RL-specific plumbing.

Direction of integration: Harbor is the primary product; Arctic ships a plugin for it. Harbor discovers the plugin via `importlib.metadata` entry points registered on the top-level `pyproject.toml`:

```
$ harbor plugins list
┃ Name                ┃ Import path                                             ┃
│ arctic-cortex-agent │ arctic_platform.integrations.harbor.agent:CortexRLAgent │
│ arctic-cortex-env   │ arctic_platform.integrations.harbor.env:HostEnvironment │
```

Every LLM call runs inside a `harbor run` trial (`BaseAgent` under `BaseEnvironment`, `RolloutDetail` written to disk), scored by Harbor's stock `Verifier` execing each task's `tests/test.sh`. Between trials, the driver reads Harbor's `result.json`, runs one Cortex GRPO step, and `sync_weights` propagates the new weights back to the same sampling sub-job — so the next `harbor run` samples from an improved model at the same endpoint. No Harbor code is modified. No custom `BaseVerifier`.

## Two sampling modes (both working E2E)

| Mode | Agent | How it reaches the sampling sub-job |
| --- | --- | --- |
| Native (default) | `CortexRLAgent` (or any agent consuming `ArcticRLClient.reconnect_config`) | RL-shaped `/generate` route, DSSST1 octet wire. |
| OpenAI-compat (`--sampling-api-base`) | Any Harbor `BaseAgent` that speaks OpenAI-chat — Terminus 2, LangChain-backed agents, custom `BaseAgent`s | `/v1/chat/completions` / `/v1/completions` / `/v1/models` on the same sub-job, translated into `ReplicaPool.generate()` in-process. |

Both routes share the same worker pool: `/generate` and `/v1/*` route through the same `ReplicaPool`, so scheduling / prefix cache / TP fan-out are identical.

### End-to-end demo: Terminus 2 training on Cortex, on-prem

```bash
# 1. Bring up the RL server (one training + one sampling GPU).
python -m arctic_platform.rl.http_server \
  --training-gpus 1 --sampling-gpus 1 --port 7000 &

# 2. Drive training with Harbor's stock Terminus 2 agent over OpenAI-compat.
harbor-cortex-train \
  --tasks-dir ./my_bench/train  --heldout-dir ./my_bench/heldout \
  --model             Qwen/Qwen3-0.6B \
  --agent             harbor.agents.terminus_2:Terminus2 \
  --sampling-api-base auto \
  --llm-backend       litellm \
  --iters 30 --prompts-per-step 8 --n-attempts 4 --lr 5e-6 \
  --out ./training-run/
```

`--sampling-api-base auto` reads the client's transport to derive the URL (on-prem: `http://localhost:7000/v1`). On Cortex, once sub-job URL routing lands at the ingress, users pass the sub-job URL explicitly — this repo does not need any further changes.

### `arctic_platform/openai_compat.py` (new; 605 LOC)

Thin OpenAI-compat router mounted on `arctic_platform.rl.http_server`.

- `/v1/chat/completions`: renders messages through the tokenizer's chat template (loaded once at `/initialize`), calls `pool.generate`, formats the response. Supports `n>1`, `stop`, `max_completion_tokens`, `presence_penalty`/`frequency_penalty`, `seed`, `stream`, `logprobs`.
- `/v1/completions`: string prompt, `list[int]` token prompt, or batched `list[list[int]]`.
- `/v1/models`: returns the currently-loaded sampling model; empty list (200, not 503) before init so LiteLLM's capability probe doesn't give up.
- `stream=True`: runs the full generation and replays as SSE deltas in OpenAI's chunked shape. Client contract is correct; first-token latency is not (needs a delta-yielding surface on `ReplicaPool` — noted in `PLAN.md`).
- Unknown OpenAI fields (`tools`, `response_format`, `parallel_tool_calls`, ...) are accepted and ignored so newer SDK versions don't 422.
- Lives above `arctic_platform.rl/` so tests can import the router without pulling the training kernel (tensordict, Ray, DeepSpeed).

### `arctic_platform/rl/http_server.py`

- Mounts `openai_compat.router` alongside `/generate`.
- `/initialize` (sampling branch) now loads the tokenizer once and stores it on `app.state.sampling_tokenizer` / `app.state.sampling_model_name`.
- `/destroy` clears the tokenizer + model_name on sampling teardown.

### Harbor plugin (`arctic_platform/integrations/harbor/`)

| File | Role |
| --- | --- |
| `models.py` | RFC data contract — `Rollout` (with `loss_mask`), `RolloutDataset`, `PostTrainingConfig`, `TrainingRun`, `InferenceEndpoint`. Mirrors Harbor's `RolloutDetail` 1:1. |
| `backend.py` | `ArcticCortexBackend` — RFC `PostTrainingBackend` protocol over `ArcticRLClient`. Owns GRPO advantage + batch construction. |
| `env.py` | `HostEnvironment(BaseEnvironment)` — host subprocesses under a per-trial root, no container. Development-only. |
| `agent.py` | `CortexRLAgent(BaseAgent)` — reference agent for the native reconnect-config path. |
| `adapter.py` | `_flatten_multi_turn` walks per-turn tokens into a flat `(prompt, completion, loss_mask)`. Marks every model-produced position across all turns as trainable; falls back to final-completion-only when the invariant fails. |
| `task_gen.py` | Arithmetic task-dir writer for the demo. |
| `train.py` | `harbor-cortex-train` CLI. Two sampling modes; `--sampling-api-base auto` derives the URL from the client's transport. |
| `aggregate.py` | `harbor-cortex-aggregate` — multi-seed aggregator with bootstrap 95% CI. |
| `PLAN.md` | Where every piece stands + what's outside this repo. |
| `RUN_LOG.md` | Transcript of the 3-seed experiment below. |

### Tests

| File | Coverage |
| --- | --- |
| `tests/openai_compat/test_openai_compat.py` | 19 unit tests. Parameter mapping (`temperature`, `top_p`, `top_k`, `seed`, `stop`, penalties, `max_completion_tokens`), `n>1` fan-out, streaming SSE structure (role → content deltas → finish frame → `[DONE]`), 503 before init, 400 for no chat template, 400 for bad body, finish-reason mapping (`length`, `abort → content_filter`, `None → stop`), token-id prompts, batched string / list-of-list-int prompts, unknown-field tolerance. |
| `tests/openai_compat/test_real_openai_sdk.py` | 4 contract tests. Spins up a live `uvicorn` subprocess and drives it with the actual `openai` Python client: chat completions non-stream + stream, completions, `models.list`. Skips cleanly when `openai` isn't installed. |
| `tests/integrations/harbor/test_adapter_multiturn.py` | 6 tests for the multi-turn flatten. |

All 29 tests pass locally.

## Experiment: 15-step GRPO on Cortex (native mode)

`Qwen/Qwen3-0.6B`, 3-digit × 2-digit multiplication (a ∈ [100, 999], b ∈ [10, 99]), 15 GRPO steps × 24 rollouts/step (6 prompts × 4 attempts), lr=5e-6, temp=0.8. Held-out 80 problems, greedy re-eval. Three seeds. Full transcript in `arctic_platform/integrations/harbor/RUN_LOG.md`.

```
                seed 0            seed 1            seed 2            n=3 aggregate
pass@1          0.362 → 0.350     0.350 → 0.400     0.375 → 0.425     Δ +0.029  95% CI [-0.013, +0.050]
mean held-out r 0.580 → 0.696     0.600 → 0.690     0.648 → 0.741     Δ +0.100  95% CI [+0.090, +0.116]
```

- pass@1 CI spans zero: a 0.6B model doesn't reliably fix 3-digit × 2-digit multiplication in 15 GRPO steps.
- Mean held-out reward moves +10.0 pp, 95% CI [+9.0, +11.6] across three seeds. On seed 0, 19 held-out problems moved out of the reward ≤ 0.05 bucket into "close but wrong" or "verbose correct".

### Non-obvious client-side bugs found and fixed during bringup

1. `ArcticRLClientConfig.training_job_id/sampling_job_id` are `Field(exclude=True)` — `model_dump_json()` drops them. Each Harbor trial subprocess was cold-starting a fresh Cortex job and blocking on `_wait_running`. Fix in `train.py`: write ids back after dump.
2. `ArcticRLClient.shutdown()` cancels the shared Cortex sub-job. With one trial per subprocess, the first trial killed the job the rest needed. Fix in `agent.py`: drop `shutdown()` from the agent; the runner cancels in its `finally`.
3. `AutoTokenizer.from_pretrained` on every trial hits HF Hub → rate limit → dropped trials. Fix: pre-warm cache in the runner, `local_files_only=True` in the agent.

## What's still outside this repo

Only one item: Cortex control-plane routing so a client can reach `https://<cortex>/sub-jobs/<sampling-sub-job-id>/v1/*` from outside. The sub-job process already answers on `/v1/*` (this PR), and on-prem the URL is just `http://<host>:<port>/v1`. Once the ingress rule is in, no code here changes.

Non-blocking follow-ups (see `PLAN.md`):

- `harbor train` subcommand upstream in Harbor — ~30-LOC Typer addition, dispatch through a `harbor.backends` entry point. `arctic_platform.integrations.harbor.train:cli` is a drop-in.
- `PostTrainingBackend.stream_progress()` — async iterator around `backend.train()`.
- Real sandbox in production runs — swap `HostEnvironment` for Harbor's `DockerEnvironment` / `ModalEnvironment` / `DaytonaEnvironment`.

## Sanitized for a public PR

Personal absolute paths stripped from `RUN_LOG.md`. Cortex sub-job UUIDs are invalidated (jobs shut down) and left in as evidence that baseline + final used the same sub-job. No PATs or internal hostnames in the diff.

## Companion RFC

`rfcs/harbor-post-training-backend.md` (working tree; not merged). This PR is the reference implementation.


sfc-gh-kganesan:
## R2E-Gym SWE agent on Cortex Training — results

Second workload on this backend, and a much harder one than the reasoning-gym run: a real black-box SWE agent (the `mini-swe-agent-plus` harness on `Qwen/Qwen3.5-4B`, multi-turn, native tool calling, 100-turn trajectories, one k3s container per rollout) against R2E-Gym tasks, trained with GRPO on Cortex. **No GPU on our side** — sampling and `fwd_bwd` both go to Cortex over HTTP, and the agent reaches the sampler through `DriverOpenAIGateway` on the cluster bridge.

Hyper-parameters are matched to the reference prime-rl run throughout. The one deliberate deviation is the loss: GRPO rather than its CISPO, since CISPO needs a Cortex-side registration.

### The loop closes

3 GRPO steps, 96 rollouts, end to end: sandboxed rollouts → R2E-graded reward → advantages → Cortex `fwd_bwd` → weight sync.

| | |
|---|---:|
| Pass rate (earned reward) | **41.7 %** (40/96) |
| Terminal-invalid traces | **74.0 %** (71/96) |
| …of which solved the task and get zeroed | 22 |
| Effective reward after override | **18.8 %** (18/96) |

### The blocker is protocol grading, not the backend

The model solves these tasks at a reasonable rate. prime-rl then zeroes any trace that trips a terminal detector, and we match that. The reference run sits at 10–23 % terminal-invalid; at our 74 %, most groups of 8 have near-zero advantage spread and there is almost nothing left to learn from.

What actually fires:

| Detector | Count | Share |
|---|---:|---:|
| `unknown_tool` | 24 | 25.0 % |
| `nonempty_content` | 19 | 19.8 % |
| output-token limit / turn budget | 12 | 12.5 % |
| `exact_repeat_no_source_edit_since` | 9 | 9.4 % |
| `invalid_tool_arguments_schema` | 9 | 9.4 % |
| `missing_tool_call` | 7 | 7.3 % |

Two of these are ours, not the model's:

- **`unknown_tool` (24)** — our task wrapper told the agent to "submit", but the harness has no `submit` tool; submission is `echo MINI_SWE_AGENT_FINAL_OUTPUT` through `execute_bash`. Fixed by passing the bare problem statement like the harness's own `get_instruction`, not yet re-measured.
- **The 12 "truncations"** all ended at exactly turn 100 — our `max_turns`, i.e. the gateway's turn budget, not the model exceeding its output-token limit.

The rest are still being split between genuine model errors and artefacts of our OpenAI-compat parsing.

### What this needed in the backend

- `openai_compat` learns **native tool calling** (Qwen3.5's XML envelope → OpenAI `tool_calls`) plus `<think>` handling for a generation prompt that opens the block and a completion that never closes it. Without it every well-formed call read as a format violation.
- **Micro-batched `fwd_bwd`** with length-sorted padding trim — 222 rollouts × 32 k tokens overflows one Cortex request — with metrics aggregated across pieces instead of reporting the last.
- **GRPO advantages honour the rollout's loss mask**, and std normalisation is configurable, so a multi-turn trajectory trains on every assistant turn.
- **`weight_sync` / `server_models` defer their `arctic_inference` imports.** `arctic_platform.rl` re-exports `weight_sync`, so an eager import dragged the GPU stack onto a CPU-only driver pod.
- Cortex HTTP errors now carry the response body, which is how the oversized `fwd_bwd` was diagnosed at all.

107 tests pass (`tests/`, excluding the `ray`/GPU-only `tests/rl` and `tests/zorro_train`).

### Not done

No convergence run. The reference curve moves 0.28 → 0.71 by step 91 of a planned 500; we have 3 steps, which says the loop closes, not that it learns. Order of work: terminal-invalid rate down first, then batch 128 and ~30 steps.

One correction worth recording: the run originally cited as the target turned out to use the **coco** harness, not `mini-swe-agent-plus`, so it is the wrong baseline for this recipe. The right comparison reaches reward 0.652 / eviction-adjusted 0.835 at step 500. Two of its settings differ from the coco run and we match the `mini-swe-agent-plus` one: `std_normalization` true, and a linear length penalty — both still outstanding here.

Details in `arctic_platform/integrations/harbor/R2E_SWE_RUN.md`; driver in `arctic_platform/integrations/harbor/examples/r2e_swe/`.

PR #71, opened 2026-08-17
Route DeepSpeed model load through ModelSpec patches

sfc-gh-truwase:
## Summary
- Register `zorro_train` and `gradient_checkpointing` in the model patch registry (alongside `liger`) and apply them via `build_model(ModelSpec)`.
- Bridge flat verl `ds_worker_config` into `ModelSpec.from_ds_worker_config` (require `attn_implementation`; ZoRRo disabled → `None`; worker bridge defaults GC-on).
- Replace inline HF load / liger / ZoRRo / post-init GC in `DeepSpeedWorker.initialize` with the factory path.
- Guard ZoRRo train patch to Qwen3-family `model_type` values only.

## Test plan
- [x] `pytest tests/model/test_model_factory.py` (22 passed, incl. attn required + non-Qwen3 reject)
- [x] CI formatting / unit-tests / semgrep green (`ad5e5b7`)
- [x] GSM8K Ray 1-step smoke after refactor (`job-20260817T165402Z`, exit 0)
- [x] GSM8K Ray 40-step quality smoke (`job-20260817T182704Z`, reached step 40)
- [x] GSM8K HTTP 10-step smoke (`job-20260817T205758Z`, `protocol=http`, reached step 10)
- [x] txt2sql BIRD 1-step smoke: Qwen3-1.7B on 2× H200, ZoRRo train on (`job-20260817T220459Z`, reward mean ~0.53)
- [x] txt2sql BIRD official recipe: Qwen3-32B on 8× H200, ZoRRo train+inference, 10 steps (`job-20260818T144500Z`, ~3h, reward mean ~0.53–0.66)

### DIFF
--- a/arctic_platform/model/config.py
+++ b/arctic_platform/model/config.py
@@ -78,31 +78,48 @@ class ModelSpec(BaseModel):
 
     @classmethod
     def from_ds_worker_config(cls, model_name: str, ds_worker_config: dict) -> "ModelSpec":
-        """Map flat ``ds_worker_config`` to ModelSpec (FA2 / bf16 / GC-on defaults)."""
+        """Transitional bridge: map flat verl ``ds_worker_config`` into a ``ModelSpec``.
+
+        Callers (adapter / SFT demos) own the knobs on the flat dict today. Longer-term,
+        ``ModelSpec`` should be constructed upstream as part of the main client config
+        instead of being inferred here.
+        """
         cfg = ds_worker_config or {}
 
-        zorro_train = None
-        if cfg.get("zorro_train_enable", False):
-            zorro_train = ZorroTrainPatch(
-                response_len=cfg.get("response_len"),
-                max_token_len=cfg.get("max_token_len"),
-                rollout_n=cfg.get("rollout_n"),
-                temperature=cfg.get("temperature"),
-                use_unpad=cfg.get("use_unpad", True),
-                world_size=cfg.get("world_size"),
-                logits_optimization=cfg.get("logits_optimization", "none"),
-                logits_optimization_peak_mem_size_in_gib=cfg.get("logits_optimization_peak_mem_size_in_gib", 4),
-                logits_compute_from_fp32_inputs=cfg.get("logits_compute_from_fp32_inputs", False),
-                logits_compute_in_fp32=cfg.get("logits_compute_in_fp32", False),
+        # Require an explicit attention backend. ZoRRo Train needs a flash-attention
+        # implementation; do not invent a second default here (ModelSpec leaves it None).
+        if "attn_implementation" not in cfg or cfg["attn_implementation"] is None:
+            raise ValueError(
+                "from_ds_worker_config requires attn_implementation "
+                "(ZoRRo Train needs a flash-attention backend)."
             )
 
+        zorro_train_patch = None
+        if cfg.get("zorro_train_enable", False):
+            # Only forward keys present in cfg; ZorroTrainPatch pydantic defaults fill the rest.
+            zorro_keys = (
+                "response_len",
+                "max_token_len",
+                "rollout_n",
+                "temperature",
+                "use_unpad",
+                "world_size",
+                "logits_optimization",
+                "logits_optimization_peak_mem_size_in_gib",
+                "logits_compute_from_fp32_inputs",
+                "logits_compute_in_fp32",
+            )
+            zorro_train_patch = ZorroTrainPatch(**{k: cfg[k] for k in zorro_keys if k in cfg})
+
+        # Worker bridge defaults GC on (historical DeepSpeedWorker behavior). Generic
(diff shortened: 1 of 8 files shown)


arctic_platform/model/config.py
```diff
@@ -57,6 +76,37 @@ class ModelSpec(BaseModel):
...
+        if cfg.get("zorro_train_enable", False):
+            zorro_train = ZorroTrainPatch(
+                response_len=cfg.get("response_len"),
+                max_token_len=cfg.get("max_token_len"),
+                rollout_n=cfg.get("rollout_n"),
+                temperature=cfg.get("temperature"),
+                use_unpad=cfg.get("use_unpad", True),
+                world_size=cfg.get("world_size"),
+                logits_optimization=cfg.get("logits_optimization", "none"),
+                logits_optimization_peak_mem_size_in_gib=cfg.get("logits_optimization_peak_mem_size_in_gib", 4),
```
sfc-gh-mwyatt: We should avoid defining defaults in more than 1 place (default of 4 is already present in the pydantic model). Same to the other fields here.

arctic_platform/model/config.py
```diff
@@ -57,6 +76,37 @@ class ModelSpec(BaseModel):
...
+                logits_compute_from_fp32_inputs=cfg.get("logits_compute_from_fp32_inputs", False),
+                logits_compute_in_fp32=cfg.get("logits_compute_in_fp32", False),
+            )
+
+        return cls(
+            model_path_or_name=model_name,
+            dtype="bfloat16",
+            attn_implementation=cfg.get("attn_implementation", "flash_attention_2"),
```
sfc-gh-mwyatt: If "flash_attention_2" should be the default, then we should define it in the pydantic model field

arctic_platform/model/config.py
```diff
@@ -57,6 +76,37 @@ class ModelSpec(BaseModel):
     )
     patches: Patches = Field(default_factory=Patches, description="Post-load patches.")
 
+    @classmethod
+    def from_ds_worker_config(cls, model_name: str, ds_worker_config: dict) -> "ModelSpec":
```
sfc-gh-mwyatt, line 80: Is this method intended just as a bridge? Ideally we would have constructed the modelspec as part of the main client config.

arctic_platform/model/config.py
```diff
@@ -57,6 +76,37 @@ class ModelSpec(BaseModel):
...
+                logits_compute_in_fp32=cfg.get("logits_compute_in_fp32", False),
+            )
+
+        return cls(
+            model_path_or_name=model_name,
+            dtype="bfloat16",
+            attn_implementation=cfg.get("attn_implementation", "flash_attention_2"),
+            patches=Patches(
+                liger=cfg.get("use_liger", False),
+                zorro_train=zorro_train,
```
sfc-gh-sbekman: I just find it a bit odd that all keys to Patches is key=bool_val, but zero is an object

at the very least `zorro_train=zorro_train` should be disambiguated to what `zorro_train` value stands for, I thought it was a `bool` but it's not.

sfc-gh-sbekman (commented):
overall looks good, with one nit below

arctic_platform/model/config.py
sfc-gh-sbekman: hmm, probably there should be no default but it should assert if it's not set.

zorro needs some version of FA

arctic_platform/model/config.py
sfc-gh-truwase: Fixed in 9cec8c5 — `from_ds_worker_config` now only forwards ZoRRo keys present in `ds_worker_config`; omitted fields use `ZorroTrainPatch` pydantic defaults as the single source of truth.

arctic_platform/model/config.py
sfc-gh-truwase: Went with @sfc-gh-sbekman’s suggestion in 9cec8c5: no silent FA2 default in the bridge. `from_ds_worker_config` now requires `attn_implementation` explicitly (`ModelSpec.attn_implementation` stays `None` for direct construction). Adapter/SFT callers already set it.

arctic_platform/model/config.py
sfc-gh-truwase: Yes — intentional transitional bridge for the flat verl `ds_worker_config` wire. Docstring updated in 9cec8c5 to say so; migrating `ModelSpec` into the main client config is follow-up (out of scope here so we don’t change the client schema in this PR).

arctic_platform/model/config.py
sfc-gh-truwase: Agreed — renamed the local to `zorro_train_patch` in 9cec8c5 so it’s clear this is a `ZorroTrainPatch | None`, not a bool like `liger` / `gradient_checkpointing`.

sfc-gh-sbekman (approved):

sfc-gh-mwyatt (approved):

PR #75, opened 2026-08-18
Refactor install to allow for minimal client-only install

sfc-gh-mwyatt:
- Add lightweight Cortex-only installation support.
- Split SFT, RL, on-prem, and verl dependencies into optional extras.
- Provide actionable install guidance when optional dependencies are missing.
- Update CI and documentation for Python 3.12 and the new install paths.

### DIFF
--- a/arctic_platform/_dependency_groups.py
+++ b/arctic_platform/_dependency_groups.py
@@ -12,12 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Report which extra to install when a subpackage's dependencies are absent.
+"""Report which dependency group to install when dependencies are absent.
 
 The base install carries config models only; each backend lives behind an extra
-(see ``pyproject.toml``). Subpackages call `require_extra` so an import under,
-say, ``arctic_platform.common`` names the extra to install instead of failing
-with a bare ``ModuleNotFoundError`` on whichever dependency happened to be first.
+(see ``pyproject.toml``). Subpackages call `require_any_dep_group` so an import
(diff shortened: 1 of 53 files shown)


arctic_platform/_dependency_groups.py
```diff
@@ -0,0 +1,90 @@
...
+        return False
+    return True
+
+
+def _missing(extra: str) -> list[str]:
+    return sorted(name for name in _provided_by(extra) if not _installed(name))
+
+
+def require_extra(*extras: str) -> None:
+    """Raise unless at least one of ``extras`` is fully installed.
```
sfc-gh-sbekman, line 82: I don't understand the contract here, why "at least one", how is the caller aware of that?

sfc-gh-sbekman: should it not say "raise unless all requirements are there"? or "raise if it at least one isn't there"?

sfc-gh-mwyatt: For some code, we can enter the path with either `rl` or `sft` (e.g., in `arctic_platform.common`). For this reason, we accept if either `rl` or `sft` dependencies are met.

sfc-gh-sbekman: could the function name reflect that? `require_either_or` or some such so that it is self-documenting?

sfc-gh-sbekman: `require_at_least_one_of('rl', 'sft')`?

sfc-gh-mwyatt: Fixed in latest commit. Now it's `require_any_dep_group`

sfc-gh-sbekman (approved):

sfc-gh-truwase (approved):

PR #77, opened 2026-08-18
Switch to `onprem` and `remote` as backend types in client config

sfc-gh-mwyatt:


arctic_platform/client/examples/sft_example.py
```diff
@@ -13,16 +13,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Unified SFT training example across the on-prem and Cortex backends.
+"""Unified SFT training example across the on-prem and remote backends.
 
     python arctic_platform/client/examples/sft_example.py --backend onprem-http
     python arctic_platform/client/examples/sft_example.py --backend onprem-ray
-    CORTEX_PAT=... python arctic_platform/client/examples/sft_example.py --backend cortex
+    CORTEX_PAT=... python arctic_platform/client/examples/sft_example.py --backend remote-cortex
```
sfc-gh-sbekman, line 20: this is ambiguous/confusing since later you use a separate `backend` and `comm_protocol` split entries, so the above `--backend` overloads the core `backend` may be don't call it `--backend` in the example?

sfc-gh-sbekman: I see actually this issue is in many places below and exists before this PR

2 related but different components are named `backend` but serve different purposes and use different set of values

sfc-gh-sbekman (commented):
there is an issue that is before this PR, please see my comments below

arctic_platform/client/examples/sft_example.py
sfc-gh-mwyatt: I plan to remove these examples soon and replace with proper recipes. This was mainly for testing during early dev work. I think we can safely ignore this problem.

sfc-gh-sbekman (approved):

sfc-gh-truwase (approved):

arctic_platform/client/config.py
```diff
@@ -21,7 +21,7 @@
     ├── training_gpus / sampling_gpus / log_prob_gpus
     ├── training     # DeepSpeed engine (ds_config owns optimizer/scheduler/batch) + checkpoint
     ├── sampling     # sampling / log-prob engines (vllm, log_prob_engine, ...)
-    └── backend_config: OnPremConfig | CortexConfig  # connection / deployment only
+    └── backend: OnPremConfig | CortexConfig  # connection / deployment only
```
sfc-gh-sbekman, line 24: is this needing updating? cortex is not longer the backend, should be RemoteConfig, no?

sfc-gh-mwyatt: Given there is only 1 remote config (cortex), I opted to keep it as `CortexConfig`

PR #81, opened 2026-08-20
[bug fix] failed training initialize leaks DeepSpeed workers

sfc-gh-sbekman:
## Summary

Training `/initialize` spawned DeepSpeed workers, then asserted `checkpoint_path` before registering the job. A missing path 500'd with workers still set, so destroy was 404 and the next initialize was 409. HTTP and Ray now reject a missing path before spawn. Spawn+init+destroy-on-failure lives in one helper both servers call.

## Testing

- Repro failed on current `main` (`AssertionError` after workers were assigned). After the fix, `tests/common/test_initialize_cleanup.py` — 4 passed, and `tests/common` — 18 passed, on `stas-dev-2-0` (`conda` `dev`).

### DIFF
--- a/arctic_platform/common/ray_server.py
+++ b/arctic_platform/common/ray_server.py
@@ -348,6 +348,7 @@ class ArcticRLRayServerState(ArcticRLServerState):
             master_port = (
                 self.ds_master_port if self.ds_master_port is not None else int(os.environ.get("MASTER_PORT", 29500))
             )
+
             def actor_options(rank):
                 if colocate and placement:
                     return _pg_options(bundle_index=rank, fraction_key="training")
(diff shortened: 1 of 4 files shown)


arctic_platform/common/ray_server.py
```diff
@@ -340,23 +341,22 @@ def _pg_options(bundle_index: int, fraction_key: str) -> dict:
...
-                    opts = dict(num_gpus=1)
-                w = DeepSpeedWorker.options(**opts).remote(rank, gpus, master_port)
-                workers.append(w)
-            master_addr = await workers[0].get_ip.remote()
-            await asyncio.gather(*[w.initialize.remote(master_addr, config_dict) for w in workers])
-            self.training_workers = workers
+                    return _pg_options(bundle_index=rank, fraction_key="training")
+                return dict(num_gpus=1)
+
+            self.training_workers = await spawn_and_initialize_workers(
```
sfc-gh-truwase, line 357: We need a separate pass to refactor the commonalities of `ray_` and `http_` servers.

sfc-gh-truwase (approved):

arctic_platform/common/ray_server.py
sfc-gh-sbekman: do we need to wait for Mike's work to finish or is this something I can work on already now?

PR #83, opened 2026-08-20
Client base class for RL & SFT clients

sfc-gh-mwyatt:
Unifies the SFT and RL clients under one shared hierarchy. Runtime behavior is largely unchanged; the main changes are API consolidation and code movement.

### Client hierarchy

```text
_ArcticClientCore              # transport, jobs, reconnect config, server state
├── ArcticClient               # blocking shared operation surface
│   ├── ArcticSFTClient        # SFT loss defaults and SFT-compatible signatures
│   └── ArcticRLClient         # adds log_probs
└── AsyncArcticClient          # awaitable shared operation surface
    └── AsyncArcticRLClient    # adds log_probs
```

Unqualified client names are blocking; clients prefixed with `Async` are awaitable.

### Changes

- Renames `ArcticRLClientConfig` to the shared `ArcticClientConfig`.
- Replaces the flat SFT configuration with the shared nested configuration.
- Moves all operation-to-`Request` builders into `client/requests.py`.
- Splits the previous client module into focused base, SFT, and RL modules.
- Preserves SFT’s `fwd_no_grad(..., processing=...)` and path-first `save_checkpoint(...)` contracts.
- Moves lightweight SFT profiling outside `common.utils` so Cortex clients do not acquire on-prem dependencies.

### Intentional API changes

- `ArcticRLClient` is now blocking; use `AsyncArcticRLClient` for async calls.
- Client factory functions and old compatibility modules are removed.
- SFT now uses the shared nested config shape and the shared `max_seq_len=8192` default.

Operation routing, transport behavior, reconnect handling, verl integration, and weight synchronization remain unchanged.


sfc-gh-sbekman (approved):
Thank you for the massive refactor, Mike!

sfc-gh-truwase (approved):

Issue #35, opened 2026-07-04
[integrations] Land verl RemoteBackend adapter under arctic_platform/integrations/verl/


sfc-gh-kganesan:
## Summary

Land the ArcticRL ⇄ verl integration as an `arctic_platform/integrations/verl/` subpackage, plugged into verl core via the `VERL_USE_EXTERNAL_MODULES` hook. This unblocks the currently-stalled [verl-project/verl#6422](https://github.com/verl-project/verl/pull/6422) by moving all Arctic-specific code out of the verl source tree while keeping the generic `RemoteBackend` abstraction in verl core.

**Target UX:**

```bash
pip install arctic-platform[verl]
export VERL_USE_EXTERNAL_MODULES=arctic_platform.integrations.verl.register
verl train ... trainer.remote_backend=arctic
```

Three commands. Zero verl fork required. Snowflake owns both the RL runtime (`arctic_platform.rl`) and the verl adapter in one repo.

## Motivation

- **PR #6422 is stuck.** Approved by `@zw0610` on 2026-06-01, then on 2026-06-30 `@wuxibin89` requested (a) port to the V1 trainer at `verl/trainer/ppo/v1/trainer_base.py`, and (b) *"keep `remote_backend` abstraction in verl and separate all `arctic_rl` implementation to external module. All specific remote backend implementation should be dynamic plugin by hook `VERL_USE_EXTERNAL_MODULES`"*.
- **Precedent exists for framework-aware code in this repo.** `arctic_platform/rl/processors/verl_grpo.py` (~440 LOC) is already a verl-shaped loss registered via `arctic_platform.rl`'s pipeline. Formalizing this under `integrations/verl/` cleans up an implicit habit.
- **Precedent exists for optional extras.** `pyproject.toml` already ships `[rl]`, `[dev]`, `[testing]`, `[formatting]`. Adding `[verl]` is one block, no new mechanism.
- **Owning both sides in one repo removes cross-repo lag.** During PR #3 review, adapter changes were blocked waiting for `arctic_training` async signatures (commit `af1ab8d`). Same repo = same PR.

### Package layout (added under `arctic_platform/integrations/verl/`)

```
arctic_platform/
    integrations/                                       # NEW
        __init__.py
        verl/
            __init__.py
            register.py                                  # ~15 LoC entry hook
            adapter.py                                   # ← was verl/workers/remote_client/arctic_rl.py (565 LoC)
[...]

PR #84, opened 2026-08-20
TRL async-GRPO backend for Arctic RL (builds on #65)

sfc-gh-truwase:
Builds on @kashif's RFC #65 ("TRL backend for Arctic RL"). Kashif's three commits are the
foundation of this branch and are preserved verbatim with his authorship:

- `Add TRL integration sketch for the Arctic RL backend`
- `Rename to forward_no_grad and forward_backward, explain the two forwards`
- `Follow TRL's fused forward_backward taking a loss callable`

This PR carries that RFC to a working, validated end-to-end GSM8K GRPO run on the Arctic
backend and adds the pieces the sketch left open.

## What this adds on top of #65

- **Rollout worker** (`integrations/trl/rollout.py`) implementing TRL's `RolloutWorkerProtocol`:
  single-turn generation + synchronous `reward_funcs` via `SyncArcticRLClient.generate`, run in a
  daemon thread of rank-0 (no CUDA in the client process).
- **Weight transfer** (`integrations/trl/weights.py`) implementing `WeightTransferProtocol`: delegates
  the train→sampler sync to `client.sync_weights` (the trainer's local params are not the source of
  truth), auto-selecting CUDA-IPC under colocation.
- **verl-style `old_log_probs` from the training engine** (`engine_old_log_probs`): `old` and the
  trainer's `new` log-probs come from the same subsystem, so the on-policy ratio sits at ~1.
- **FA2-correct varlen packing**: the server packs sequences varlen-style (per-sequence `position_ids`
  resets). Only a flash-attention backend keeps them separated via block-diagonal `cu_seqlens`; `sdpa`
  attends across packed boundaries and silently corrupts per-token log probs (the cause of the
  ratio/KL explosion). `ModelSpec.from_ds_worker_config` now defaults `attn_implementation` to
  `flash_attention_2` instead of raising.
- **vLLM 0.26 support**: `transformers>=5.2`; ported ArcticInference server layer to vLLM 0.26 lives in
  the companion `ArcticInference` tree.
- **Thread-safe Ray transport** for the in-process Ray client.
- **36 CPU unit tests** (`tests/integrations/trl/`, `tests/model/`) covering the batch-layout
  conversions, the roll(-1)→current-token log-prob frame, weight-transfer routing, and the
  `attn_implementation` default.

## Test plan

- [x] `pytest tests/integrations/trl tests/model/test_model_factory.py::TestFromDsWorkerConfig` — 36 passed (CPU).
- [x] End-to-end GSM8K GRPO on the Arctic backend, 100 steps, vLLM 0.26 (HTTP + Ray transports): E2E PASSED, ratio≈1 / kl≈0.
- [x] Multi-seed (7/42/123) baseline-vs-Arctic reward comparison on vLLM 0.26.

Note: this branch is rebased onto current `main`, so #65's commits appear here rebased (same author);
merging this can supersede #65, or #65 can merge first and this rebase cleanly on top.

## BIRD / text2sql — native TRL vs Arctic RL + ZoRRo

- **Native TRL async-GRPO** — in-process trainers + `vllm serve` (the stock TRL path).
- **Arctic RL + ZoRRo + load balancer** — disaggregated Arctic backend with prompt-prefix ZoRRo and load-balanced microbatches.

SQL reward is execution match on BIRD. Last-20 over 50 steps is the quality number. Train-step is `step_time` mean of optimizer step 2+.

### Qwen3-8B (1 node, 4 train + 4 sample, 50 steps)

GBS 2048, n=16, seq 36864 / completion 4096. Native used microbatch 4 / GAS 128 (microbatch 16 OOM). Arctic + ZoRRo used microbatch 16 / GAS 32.

| | Native TRL async-GRPO | Arctic RL + ZoRRo + LB |
| --- | --- | --- |
| `train_runtime` | 88931 s (24.7 h) | 46250 s (12.8 h), **1.92×** |
| Iter / train-step / idle (step 2+) | 1752 / 1104 / 620 s | 880 / 286 / 579 s |
| Last-20 SQL reward | 0.673 | 0.702 (**+0.029**) |
| Mean / end reward | 0.653 / 0.682 | 0.675 / 0.718 |

Train-step is **3.86×**; E2E is **1.92×** because generate still costs ~579 s of trainer idle on the Arctic path.

Caveats: 8B microbatch sizes differ (4 vs 16). HTTP rollout order still differs, so per-step train `reward` is a noisy A/B; last-20 over 50 steps is the quality gate.


sfc-gh-truwase:
@kashif FYI

arctic_platform/integrations/trl/client.py
```diff
@@ -0,0 +1,323 @@
...
+    pad_token_id: int,
+    max_token_len_per_gpu: int,
+    calculate_entropy: bool,
+    logits_optimization: str = "none",
+    logits_optimization_peak_mem_size_in_gib: int = 4,
+    logits_compute_in_fp32: bool = False,
+) -> dict:
+    """``meta`` for ``run_pipeline`` (same keys as verl-GRPO ``base_meta``)."""
+    return dict(
+        zorro_train_enable=False,
```
sfc-gh-sbekman: should we not have it on by default?

arctic_platform/integrations/trl/client.py
```diff
@@ -0,0 +1,323 @@
...
+        self.last_grad_norm = norm[0] if isinstance(norm, list) else norm
+
+    def zero_grad(self, set_to_none: bool = True) -> None:  # type: ignore[override]
+        pass  # server clears grads in step()
+
+
+# Packed [1, T] <-> padded [B, S]. Advantages / old_log_probs stay in TRL.
+
+
+def _segment_lengths(position_ids: torch.Tensor) -> torch.Tensor:
```
sfc-gh-sbekman: unintentional gap?

arctic_platform/integrations/trl/loss.py
```diff
@@ -0,0 +1,42 @@
...
+
+from typing import Any
+
+import torch
+
+from arctic_platform.rl.processors.pipeline import register_loss_fn
+
+
+@register_loss_fn("weighted_logprob_sum")
+def weighted_logprob_sum(
```
sfc-gh-sbekman: should this be generic and available to other integrations? why specific to TRL and require code replication?

arctic_platform/model/config.py
```diff
@@ -87,12 +87,13 @@ def from_ds_worker_config(cls, model_name: str, ds_worker_config: dict) -> "Mode
         """
         cfg = ds_worker_config or {}
 
-        # Require an explicit attention backend. ZoRRo Train needs a flash-attention
-        # implementation; do not invent a second default here (ModelSpec leaves it None).
-        if "attn_implementation" not in cfg or cfg["attn_implementation"] is None:
-            raise ValueError(
-                "from_ds_worker_config requires attn_implementation (ZoRRo Train needs a flash-attention backend)."
-            )
+        # Default to flash_attention_2 when unset. The training server packs sequences varlen-style (multiple
```
sfc-gh-sbekman: Haven't we just removed this default elsewhere? 

I don't think fa2 is a sensible default, we have no idea what is installed and given we will be on at least h200, fa2 is a bad choice, need fa3 at least and fa4 on blackwell.

sfc-gh-sbekman: We should assert if the client hasn't provided this setting.

what we do require is that it's some sort of FA, but not v2 in particular.

sfc-gh-sbekman (changes requested):
How much of this code is TRL specific, and if a lot of it might be re-used by other integrations, no, and if so should it then live under integrations/trl?

sfc-gh-truwase:
## TRL ↔ Arctic RL integration — achievements summary

Milestone recap of the TRL async-GRPO ↔ Arctic RL work. Foundation is @kashif's RFC #65 (three commits preserved verbatim with his authorship); this branch carries it to a working, validated backend.

### Landed in this PR (#84)
- **`ArcticTrainingClient`** bridging TRL's `AsyncGRPOTrainer` to the Arctic RL server (DeepSpeed + vLLM) — implements TRL's `forward_backward` contract (packed `[1,T]` varlen ↔ server `[B,S]`, log-probs/entropy mapped back to TRL's shifted frame).
- **Rollout worker** + **weight transfer** protocols; client process is **CPU-only, no CUDA, no DeepSpeed** (honors Arctic's client design).
- **On-policy correctness**: `old_log_probs` sourced from the training engine so the ratio sits at ≈1 (fixed the sampler/trainer misalignment).
- **FA2-correct varlen packing** (block-diagonal `cu_seqlens`); `ModelSpec` now defaults `attn_implementation=flash_attention_2` instead of raising.
- **vLLM 0.26 stack** (torch 2.11 / transformers 5.x / DeepSpeed 0.19.5) + thread-safe in-process **Ray** transport (and HTTP).
- **CPU unit tests** for the batch-layout conversions, log-prob frame, weight routing, and attn default.
- **Validated (Qwen3-1.7B, GSM8K, vLLM 0.26, 100 steps)**: on-policy invariant matches the native TRL baseline (ratio≈1, kl≈0); reward converges to the baseline by steps 81–100 (window gap ≈ 0).
- **Server-side fused loss**: recovers GRPO loss ingredients from TRL's `loss_fn` **closure introspection** and runs forward+loss+backward in one fused server call (removes the extra client round-trip). *(loss.py / client.py + `test_server_side_loss.py`, `test_trl_grpo_loss.py`)*
- **ZoRRo train + `load_balancer`**: verl-style structured batches for prompt-prefix dedup; enabled the load balancer (previously crashed) by fixing the group-tiling invariant (`per_device_bsz` = `num_generations × training_gpus`) and the deduped-group token budget, plus a **client-side fail-fast** for ragged groups. *(client.py + `test_zorro_batch.py`)*
  - **Measured (30-step BIRD)**: LB is crash-free and perfectly on-policy (ratio 1.0 / kl 0.002), with ~22% faster fwd+backward from prefix dedup.
- **BIRD (txt2sql) task**: shared data loader + SQL exec-match reward (lives in the runner repo, CPU unit-tested).

_Reviewed with Grok + Codex; findings applied._

arctic_platform/integrations/trl/loss.py
kashif: Agreed, this one should be generic. It is `(logprobs * weights).sum()` with nothing TRL in it, and any integration doing a client side loss needs exactly it.

By the rule you applied in #35, the verl shaped loss moved under `integrations/verl` while plain `grpo` stayed in `rl/processors`. This is not TRL shaped so it goes the other way. `trl_grpo` below is TRL shaped and belongs here.

Not hypothetical either: SkyRL's `integrations/arctic_rl` passes `loss_fn="verl_grpo"` today, so one integration already leans on another integration's loss. A generic surrogate is what lets that stop.

kashif:
@sfc-gh-sbekman on the broader question, my read of what is actually TRL coupled:

- `client.py` knows TRL's packed row format and implements its protocol, so TRL specific.
- `rollout.py` and `weights.py` implement TRL protocols, thin shims over generic capabilities.
- `loss.py` splits: `trl_grpo` is TRL shaped, `weighted_logprob_sum` is not. Replied on that one inline.

So the only piece I would move out is the generic surrogate.

kashif:
@sfc-gh-truwase on the two things you asked for.

**Ownership.** TRL owns the protocol, the in process client, the trainer seam, and a conformance suite. You own the adapter, rollout, weight transfer and the losses. Same split you already have on the verl side.

The one thing I think we should take that we do not have yet is documenting and freezing the packed batch format. Every remote backend writes the same padding free conversion, so leaving it undocumented just moves the N by M problem from losses to layout. I will open that separately.

**Correctness.** My one real concern is the closure introspection in the server side loss. The guards are good, but the case they cannot catch is a user who copies `compute_loss` and changes the objective: the freevar names still match, the ingredients extract cleanly, and the server runs stock GRPO instead of their loss. Wrong gradients, no error. Since ZoRRo needs `server_side_loss=True`, that is also the path with the 22% win, so it is the one people will actually run. And if it becomes the default, as suggested above, it stops being an opt in hazard.

Cheaper fix is to let us declare it rather than you infer it: we attach an explicit spec to the closure when it is the stock GRPO loss, and you take the fast path only when that spec is present. A user loss never carries it so it falls back on its own. One line on our side, and we can test it, so a break shows up in our CI instead of in a run.

Worth saying the callable path is not just our fallback: tinker runs it in production for non GRPO work, `train_dpo.py` passes `dpo_loss_fn` straight into `forward_backward_custom`. And on the N by M point, prime-rl carries its own set (`default_loss_fn`, `ipo_loss_fn`, `ref_kl_loss_fn`, `ce_loss_fn`), verl has 12, TRL has 8.

**Perf.** I have not measured anything myself, so I am taking the 22% and the ratio/kl numbers at face value.

sfc-gh-truwase:
> **Perf.** I have not measured anything myself, so I am taking the 22% and the ratio/kl numbers at face value.

Please ignore perf at this time. This was an inadvertent leak.

sfc-gh-kganesan:
Re: the client-side loss mechanism — this now works end to end on Cortex, and it needs **no new server-side loss function and no image rebuild**.

The key observation is that `grpo` is already a general weighted-log-prob surrogate. Its gradient with respect to log-probs at ratio 1 is `-advantages / batch_num_tokens`, so if you pin `old_log_probs_shifted` to the log-probs `forward` just returned (making the ratio exactly 1) and set `advantages = -w * batch_num_tokens`, the server applies exactly `w = dL/dlogprobs`. Registering `weighted_logprob_sum` server-side is therefore a convenience, not a prerequisite — nothing about the demo has to wait on a Jenkins build.

Verified at two levels:

**Offline**, against the image's real `_grpo_loss`, comparing a direct backward of the true loss with the two-pass client-side path. Parameter gradients agree to **3.3e-16**, with and without masking.

**On a live Cortex job** (qa6 `dsa-test`, Qwen3-8B, 2xH200), running this PR's `trl_grpo` as the client-side loss with every hop going through Arctic-Platform's `CortexTransport`:

```
direct=99.02234650  surrogate=99.02126312  rel.diff=1.094e-05
PASS: TRL loss client-side, Arctic-Platform transport, Cortex backend
```

The comparison is self-verifying — on identical weights, the surrogate's `grad_norm` has to match a direct server-side `grpo` call carrying the real advantages. Server metrics corroborate the mechanism: the surrogate call reports `importance_weight=1.0000001` and `approx_kl=6.8e-08`, i.e. the ratio really is pinned to 1, while the direct call reports `1.0513` (= e^0.05) for the deliberately offset behavioural policy.

Also worth recording: `trl_grpo` and `arctic_training`'s `grpo` produce an identical client-side loss value (`0.20239189`) at `grad_accum_steps=1`, which is what you would expect given the normalizations coincide there.

Two dependencies this needed, both now verified on the debug image: dss-platform#410 (the zone must un-base64 the `/operation` envelope before parsing the frame) and dss-client#83 (chunking). Arctic-Platform#95 carries the matching transport change — `forward` had to move off the octet route onto `/operation`, because `/{job}/forward` is not reachable through the control plane.

What is still genuinely blocked is TRL's `AsyncGRPOTrainer` *driving* the loop, which needs trl#6676's `training_client=` hook. The loss mechanism underneath it is no longer the blocker.

arctic_platform/integrations/trl/client.py
sfc-gh-truwase: Left opt-in. ZoRRo needs `server_side_loss=True`, a fixed `response_len`, and a flash-attention backend; defaulting it on would also take the closure-introspection fast path that @kashif flagged. Recipes turn it on explicitly (`ARCTIC_ZORRO=1`).

arctic_platform/integrations/trl/client.py
sfc-gh-truwase: Not a missing helper — the extra blank was just spacing. Client-side loss still keeps advantages / old_log_probs inside TRL's `loss_fn`. The later `server_side_loss` path extracts them from the closure and ships them on the fused `fwd_bwd`. Comment updated to say that.

arctic_platform/integrations/trl/loss.py
sfc-gh-truwase: Moved. `weighted_logprob_sum` now lives at `arctic_platform.rl.processors.weighted_logprob` and is registered with the other generic losses. `trl_grpo` stays here. The old `integrations.trl.loss.weighted_logprob_sum` import still re-exports it.

arctic_platform/model/config.py
sfc-gh-truwase: Reverted. `from_ds_worker_config` requires an explicit `attn_implementation` again and does not default to FA2. FA2 / FA3 / FA4 are all accepted; the client picks the version that matches the GPU and installed kernels.

arctic_platform/integrations/trl/client.py
sfc-gh-truwase: Left opt-in. ZoRRo needs `server_side_loss=True`, a fixed `response_len`, and a flash-attention backend; defaulting it on would also take the closure-introspection fast path that @kashif flagged. Recipes turn it on explicitly (`ARCTIC_ZORRO=1`).

arctic_platform/integrations/trl/client.py
sfc-gh-truwase: Not a missing helper — the extra blank was just spacing. Client-side loss still keeps advantages / old_log_probs inside TRL's `loss_fn`. The later `server_side_loss` path extracts them from the closure and ships them on the fused `fwd_bwd`. Comment updated to say that.

arctic_platform/integrations/trl/loss.py
sfc-gh-truwase: Moved. `weighted_logprob_sum` now lives at `arctic_platform.rl.processors.weighted_logprob` and is registered with the other generic losses. `trl_grpo` stays here. The old `integrations.trl.loss.weighted_logprob_sum` import still re-exports it.

arctic_platform/model/config.py
sfc-gh-truwase: Reverted. `from_ds_worker_config` requires an explicit `attn_implementation` again and does not default to FA2. FA2 / FA3 / FA4 are all accepted; the client picks the version that matches the GPU and installed kernels.

PR #86, opened 2026-08-22
Add ArcticOPDClient for on-policy distillation

sfc-gh-truwase:
## Summary
- Add `ArcticOPDClient`, a composition of two `SyncArcticRLClient`s: student train+sample and a sampling-only frozen teacher.
- Isolate concurrent local servers (disjoint Ray/DeepSpeed ports, `--no-ray-auto-attach`, per-server CUDA) so student and teacher do not share one sampling pool.
- Register `on_policy_distill` reverse-KL loss, accept token-id prompts for exact teacher scoring, and pass `weight_format=hf` on Cortex student weight sync.
- Document the client in `docs/opd.md` and add `score_teacher_topk` / `fwd_no_grad` for a later TRL-style async distillation training client (#106).

Follow-up (not in this PR): #106 adds the CPU-only `ArcticOPDTrainingClient` for `AsyncDistillationTrainer`.

### CPU
- [x] `pytest tests/opd` (19 passed): loss, teacher scoring, composed client ops (student/teacher generate, fwd_bwd, step, student-only weight sync)

### Live GPU (on-prem, FA2 trainer, composed client)
Same recipe unless noted: 80 steps, warmup 20, lr `1e-6`, max seq 512, gen 32, 8 rollouts, student train+sample on GPU 0, teacher sample on GPU 1.

```bash
python -m arctic_platform.opd.examples.run_on_policy_distill \
  --live \
  --student-model Qwen/Qwen3-0.6B \
  --teacher-model Qwen/Qwen3-1.7B \
  --steps 80 --warmup-steps 20 --lr 1e-6 \
  --max-seq-len 512 --max-tokens 32 \
  --training-gpus 1 --sampling-gpus 1 --teacher-sampling-gpus 1 \
  --server-cuda-visible-devices 0 \
  --teacher-server-cuda-visible-devices 1 \
  --attn flash_attention_2
```

- [x] Qwen3-0.6B student / Qwen3-1.7B teacher, 80-step FA2: exit 0; log prints `composed student=SyncArcticRLClient(...) teacher=SyncArcticRLClient(...)`; mean distill KL ~0.64, last-10 ~0.49 (matches pre-composition FA2 80-step within noise)
- [x] Qwen3-1.7B student / Qwen3-4B teacher, 80-step FA2: exit 0; mean distill KL ~0.72, last-10 ~0.62
- [x] Confirm student weight sync does not target the teacher, and the two local servers use disjoint Ray ports / CUDA devices

### CI
- [x] Formatting, unit-tests, and semgrep green on this PR

### DIFF
--- a/arctic_platform/common/deepspeed_worker.py
+++ b/arctic_platform/common/deepspeed_worker.py
@@ -50,6 +50,9 @@ from arctic_platform.common.utils.debug import pr0
 from arctic_platform.common.utils.debug import see_memory_usage
 from arctic_platform.model import ModelSpec
 from arctic_platform.model import build_model
+from arctic_platform.model.implementations.qwen35.hf_vllm_weight_sync import apply_qwen35_sync_op
+from arctic_platform.model.implementations.qwen35.hf_vllm_weight_sync import plan_qwen35_vllm_sync
+from arctic_platform.model.implementations.qwen35.hf_vllm_weight_sync import to_vllm_sync_weights
 
 logger = logging.getLogger(__name__)
 
@@ -163,6 +166,11 @@ class DeepSpeedWorker:
         ds_worker_config = job_config.get("ds_worker_config") or {}
         ds_worker_config["world_size"] = self.world_size
         self.ds_worker_config = ds_worker_config
+        # vLLM 0.26 pulls tilelang; FLA then JIT-compiles GDN backward via
+        # cute WGMMA and dies. Force the Triton path that v4 trained with.
+        if ds_worker_config.get("fla_tilelang", True) is False:
+            os.environ["FLA_TILELANG"] = "0"
+            os.environ["FLA_DISABLE_BACKEND_DISPATCH"] = "1"
 
         # Build the DeepSpeed config per job type. Training engines get an
         # optimizer; the reference/log-prob engine is forward-only and is
@@ -181,6 +189,21 @@ class DeepSpeedWorker:
         spec = ModelSpec.from_ds_worker_config(model_name, ds_worker_config)
         loaded = build_model(spec)
         model = loaded.model
+        self._maybe_inject_fp32_lm_head(model, ds_worker_config)
+
+        # Qwen3.5 instantiates a ViT even on text-only jobs. Unused trainable
+        # params produce no grads and stall ZeRO-3 reduction; freeze them
+        # before DeepSpeed registers the param set.
+        from arctic_platform.model.implementations.qwen35.vlm import freeze_unused_vision_tower
+
+        frozen = freeze_unused_vision_tower(model, self.rank)
+        if frozen:
+            logger.info(
+                "rank=%d froze %d vision-tower params (text-only job; unused params "
+                "produce no grads and stall ZeRO-3 reduction)",
+                self.rank,
+                frozen,
+            )
 
         zorro_train_enable = ds_worker_config.get("zorro_train_enable", False)
         self.dedup_actor_model_once_patcher = getattr(model, "_arctic_zorro_once_patcher", None)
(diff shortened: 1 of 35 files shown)


arctic_platform/common/deepspeed_worker.py
```diff
@@ -50,6 +50,9 @@
 from arctic_platform.common.utils.debug import see_memory_usage
 from arctic_platform.model import ModelSpec
 from arctic_platform.model import build_model
+from arctic_platform.model.implementations.qwen35.hf_vllm_weight_sync import apply_qwen35_sync_op
```
sfc-gh-truwase, line 53: qwen3.5 model patching and ops should move to model/implementations. deepspeed_worker should not have any model-specific logic.

arctic_platform/common/deepspeed_worker.py
```diff
@@ -367,6 +390,87 @@ def _inject_sft_global_token_meta(self, loss_fn: str, batch_data, meta_data: dic
...
+            sys.modules[layers_pkg] = layers
+        spec = importlib.util.spec_from_file_location(name, qwen35 / "models" / "layers" / "lm_head.py")
+        if spec is None or spec.loader is None:
+            raise ImportError(f"cannot load {name} from {qwen35}")
+        module = importlib.util.module_from_spec(spec)
+        sys.modules[name] = module
+        spec.loader.exec_module(module)
+        return module.inject_prime_lm_head
+
+    def _maybe_inject_fp32_lm_head(self, model, ds_worker_config: dict) -> None:
```
sfc-gh-truwase, line 426: Should move to model patching pass driven by config.

arctic_platform/common/deepspeed_worker.py
```diff
@@ -367,6 +390,87 @@ def _inject_sft_global_token_meta(self, loss_fn: str, batch_data, meta_data: dic
...
+        )
+        logger.info(
+            "rank=%d injected lm_head fp32=%s chunk_size=%s fused_cross_entropy=%s",
+            self.rank,
+            want_fp32,
+            chunk_size,
+            fused_ce,
+        )
+
+    def _inject_opd_global_token_config(self, loss_fn: str, batch_data, meta_data: dict, processing: dict) -> None:
```
sfc-gh-truwase, line 450: Should be model patching pass outside.

arctic_platform/common/deepspeed_worker.py
```diff
@@ -402,14 +506,20 @@ def _forward_maybe_backward(self, batch: dict, backward: bool) -> dict:
                 pr0(f"[DeepSpeedWorker] {tag}: {k=}: {v.shape=}")
 
         grad_accum_steps = self.engine.gradient_accumulation_steps()
+        max_tokens_per_mb = self.ds_worker_config.get("max_tokens_per_mb")
```
sfc-gh-truwase, line 509: Reimplement using DS unmanaged GAS API
https://deepspeed.readthedocs.io/en/rtd-staging/training.html#unmanaged-gradient-accumulation

arctic_platform/common/deepspeed_worker.py
```diff
@@ -552,9 +667,23 @@ def forward_no_grad(self, batch: dict) -> dict:
         timers.stop_and_print_elapsed(tname)
         return results
 
-    def step(self) -> dict:
+    def _apply_learning_rate(self, learning_rate: float) -> None:
```
sfc-gh-truwase, line 670: Can this be reimplemented using DS LR scheduler

arctic_platform/common/weight_sync_extension.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 Snowflake Inc.
```
sfc-gh-truwase, line 1: Move to Arctic-Inference

sfc-gh-xyu:
## Status

`draft: true` · `mergeable_state: "dirty"` (conflicts with `main`) · zero completed check runs · no review yet other than the author's own. Flagging that first — the rest is on the code.

## Verified correct

- `avg_loss = sum(captured_losses)` (`pipeline.py:681`) is the right combination given each microbatch is normalized by the global token count. The comment on why a token-weighted mean would give `Σ(S·c)/Σ(c²)` names the actual trap.
- `did_unpad` in `run_pipeline` fixes a pre-existing bug where an already-packed batch was re-padded on the way out.
- `_gather_named_weights` adding `.detach().clone()` inside `GatheredParameters` fixes a real ZeRO-3 use-after-free; the old `get_weights` let `p.data` escape the gather context.
- `model_reads_varlen_kwargs` keys off `config.layer_types` rather than a model-name string. Correct choice.
- The loss is a faithful port: k3 with an input clamp, output cap opt-in via `kl_clamp_max`, k1 rejected as non-trainable with the gradient argument spelled out.

Confirmed end-to-end from the v16 run (`autorun/logs/job-20260902T231943Z.log`, 152 steps): `distill_batch_num_tokens == tokens/scored` and `distill_dp_size == 8` on every step, and `loss/avg == kl/per_token` with max |Δ| = 0. The global-token normalization works in the shipped configuration.

## One issue worth fixing in this PR

**Mis-scoped diagnostic metrics.** `on_policy_distill.py` emits both paired keys (`distill_kl.sum` / `distill_kl.tokens`) and unpaired ones (`distill_kl_sum`, `distill_kl_count`, `distill_kl_max`, `distill_k1_sum`, `distill_abs_delta_sum`, `teacher_logprob_sum`, `student_logprob_sum`, `sampler_train_*`). Observed in the v16 log:

```
tokens/scored / distill_kl_count = 16.0000 on all 152 steps   (n_rollouts = 16)
```

So `distill_kl_count` is a mean per-microbatch token count, not a count. By the same aggregation `distill_kl_sum` is a mean-of-microbatch-sums, `distill_kl_max` a mean-of-maxes, and `sampler_train_abs_delta_mean` an unweighted mean-of-means across rollouts spanning 130–13719 tokens in that run. (I inferred the averaging rule from these ratios; `combine_metric_microbatches` is not in this diff.)

Training correctness is unaffected — `loss/avg` is right. But these are the keys used for cross-stack comparison, and `distill_kl.sum` vs `distill_kl_sum` differ by one character with opposite semantics. Suggest pairing every diagnostic, or renaming the unpaired ones to carry their scope.

## Latent, guarded today — prefer a raise over a silent fallback

Each is currently unreachable in the shipped path; flagging so a config change doesn't turn one into a silent wrong number.

- `_resolve_distill_norm` (`:126-127`) defaults `dp_size` to 1 and allows `batch_num_tokens=None`, which makes `agg_loss` normalize locally. Harmless today because `_inject_opd_global_token_config` is called unconditionally and v16 confirms it populates. Under `sum(captured_losses)` a missing value would scale the loss by the microbatch count, so it should raise rather than default.
- `deepspeed_worker.py:465-469` all-reduces token counts over the default process group and sets `dp_size = world_size`. Correct only while DP == world size; `--student-tp 1` today, but `packing.py` already has an explicit Ulysses branch.
- `deepspeed_worker.py:516` drops the GAS-consistency check when `max_tokens_per_mb` is set. The intended invariant per the comment is GAS == 1; asserting that is safer than removing the check.
- `on_policy_distill.py:242` `nan_to_num(logprobs, nan=0.0)` maps a non-finite student logprob to p=1.0, yielding a large confident gradient. Prefer masking those positions or raising.
- `deepspeed_worker.py:171` sets `FLA_TILELANG=0` only when `fla_tilelang` is explicitly `False`, but the comment says the default tilelang path "dies". If that's accurate the default should be inverted; also `os.environ` here is a no-op if `fla` is already imported.
- `pipeline.py:239` `_ENGINE_FWD_SKIP` is a one-key blacklist, while `_run_pipeline_with_packing` now forwards every packed key to the model. A whitelist from the forward signature would fail at the boundary instead of deep in HF.
- `pipeline.py:277` `torch.roll` labels wrap across packed sequence boundaries; safe only because `loss_mask` excludes each sequence's last position. Worth a comment and a test, since the invariant lives in another file.

## Coverage

I read the loss, `packing.py`, `pipeline.py`, and `deepspeed_worker.py` — roughly 800 of 4805 added lines. I did not review the 1103-line example, `opd/{client,config,scoring}.py`, the qwen35 weight-sync map, the `http_server`/`ray_cluster` port changes, or ~1450 lines of tests. The weight-sync map deserves a careful read, since a wrong parameter mapping would corrupt the student without failing anything.

sfc-gh-truwase (commented):
## Review (Mike skill)

Base `bcea70e` (origin/main) … head `c7c2689`. Packed teacher-logprob fix is on origin. On-prem v18 152-step passed after that fix.

### Findings

1. **High — Cortex OPD `fwd_bwd` cannot deliver teacher logprobs.** `arctic_platform/opd/client.py` puts the full batch in Cortex `context` and only model keys in `kwargs`. `promote_batch_dim_to_batch` moves `loss_mask` but not `teacher_log_probs_shifted` (`BATCH_DIM_CONTEXT_KEYS` in `batch.py`). The loss reads the batch dict only. Cortex weight-sync in this PR is therefore not a working train path. On-prem HTTP (the live recipe) is unaffected.

2. **High — Vision freeze is unconditional on the shared DeepSpeed worker.** `deepspeed_worker.py` always calls `freeze_unused_vision_tower` with no text-only / OPD gate. Any Qwen3.5 VLM job on this worker will freeze `model.visual` even if pixels are in the batch.

3. **Medium — Local Ray ports collide when HTTP ports share `% 10`.** `_local_cluster_env` keys off `http_port % 10`. Validation only requires `teacher_port != student_port`. `--port 18100 --teacher-port 18110` (or any congruent pair) shares GCS / dashboard / MASTER_PORT / worker ranges. Default `port+1` (18180/18181) is fine; tests only cover that.

4. **Medium — Failed `pdsh` after `ray start --head` does not tear down the head.** `ray_cluster.py` raises on pdsh failure before `_spawned_cluster = True`, so `_shutdown()` skips cleanup. Retries leave GCS/ports occupied.

5. **Medium — PR description is stale.** Summary still says `SyncArcticRLClient`. Head uses unified `ArcticRLClient` / `ArcticClientConfig`. v18 GPU results (PR #86 `c7c2689`) are not in the test plan.

6. **Low — `distill_kl_count` comment is wrong.** `run_on_policy_distill.py` claims it is a per-rollout average; `_count` keys are summed in `combine_metric_shards`. `check_metrics` already uses it as a token count. Do not “fix” `check_metrics` off that comment.

### Residual

- `kl_clamp_max=-1` would clamp KL to a negative cap (opt-in; default is unset).
- `_hostfile_hosts` swallows `OSError` and returns empty (silent isolation miss if the hostfile path is wrong).
- `pack_n_mbs` is still not in v18 jsonl (`None`).
- DSS API spec N/A (no client-visible DSS contract change). Targeted packing UTs passed locally; full `tests/` hung on dist init here.

Packed inner-batch teacher tensors / `_engine_forward_kwargs` isolation look correct and match `test_packing_keeps_opd_teacher_logprobs_on_inner_batch`.

PR #88, opened 2026-08-24
cortex training client doc refresh

sfc-gh-jrasley:


sfc-gh-thonguyen:
Drive-by from working on the Cortex `forward` operation — filed the detail as
#98, but two notes are specific to `eval_nll` in
`cortex_training/recipes/sft/conversational/train.py`:

**1. The eval currently perturbs training** (independent of anything I'm changing).
`forward_loss` is an alias for `forward_backward`, so `eval_nll` runs a full backward
per eval batch and never steps — immediately before `forward_backward_step`. There's
no `zero_grad()` anywhere in the dss-platform training worker (only
`engine.backward()` and `engine.step()`, and DeepSpeed zeroes inside `step()`), so
those eval gradients accumulate into the next training update. Raising `eval_every`
reduces the exposure but doesn't remove it.

**2. If you move that eval to `forward`, the label convention bites.** `forward`
gathers logprobs out of `logits[t]` and never shifts, so it needs
`labels[t] == input_ids[t+1]`. With the default `model_provider="huggingface"`,
`use_next_token_labels()` returns False and `sequence_from_conversation` emits
unshifted labels — which the plain-SFT forward-backward path shifts server-side but
`forward` does not. Nothing validates the alignment, so it returns plausible-looking
wrong numbers.

The tidy fix for both is to send
`processing={"loss_fn": "causal_cross_entropy", "post": ["compute_logprobs"], "config": {}}`.
That skips the provider-dependent shift and routes the supervised loss through the
same `run_pipeline` path `forward` uses, so train and eval share one convention on
every provider — and the two losses become directly comparable, which makes the eval
self-checking.

No action needed for this PR if the eval stays on `forward_backward`; #98 has the
code references and the `forward` dependencies. Happy to send a patch once
snowflake-eng/dss-client#83 and snowflake-eng/dss-platform#410 land.

PR #93, opened 2026-08-25
Unified client recipes

sfc-gh-mwyatt:
Ports cortex-client recipes to unified client. Largely a move/refactor, with some fixes to support these recipes properly on cortex backend. Next steps will be support with onprem backend.

- **Port two standalone training recipes** (`recipes/sft/standalone/`, `recipes/rl/standalone/`) from `cortex-client` to the unified `arctic_platform.client`.
- **Fix `forward-backward` 400s on Cortex**: restore `force_chunk` in the DSSST1 encoder and send the wire's own `fwd-bwd` operation label in chunk metadata.
- **Fix `KeyError: 'request_id'` on weight sync**: Cortex's `/operations` interface have inconsistent behavior. `weight-sync` returns a `request_id` to poll, while `reset-prefix-cache` and four other ops answer inline. Dispatch on the response instead of assuming everything is async.
- **Make wake/sleep no-ops on Cortex**, which never colocates training and sampling, so shared client flows like `sync_weights()` stay backend-portable.
- **Support LoRA on Cortex**: add `TrainingConfig.peft`, forward it through `to_cortex()`, and add `weight_format` to `sync_weights()`.
- **Fail fast instead of silently dropping config**: validators reject `training.peft` and `weight_format` against on-prem, and HTTP errors now carry the server's response body.

Some of the fixes listed above were cherrypicked from #55


recipes/rl/standalone/README.md
```diff
@@ -0,0 +1,117 @@
...
+```bash
+uv pip install -e '.[cortex]'
+uv pip install -r recipes/rl/standalone/requirements.txt
+```
+
+Copy the connection template and fill in your Snowflake host and PAT. To keep the
+PAT out of the file, drop the `pat` key and export `CORTEX_PAT` instead.
+
+```bash
+cp recipes/config.json{.template,}
```
sfc-gh-sbekman: this looks borked/incomplete, no?

recipes/rl/standalone/requirements.txt
```diff
@@ -0,0 +1,18 @@
...
+# Config entrypoints (`config=... lora_rank=32` on the command line).
+chz
+
+# Dataset loading; also pulled in by tinker-cookbook.
+datasets
+
+# Renderers (chat templates + supervised weights), ml_log metric logging, and the
+# MATH environment / grading used by rl_loop.py. The [math-rl] extra carries the
+# grading dependencies.
+tinker-cookbook[math-rl] @ git+https://github.com/thinking-machines-lab/tinker-cookbook.git@nightly
```
sfc-gh-sbekman: are we sure on including a dependency on nightly that may break at any time?

recipes/sft/standalone/README.md
```diff
@@ -0,0 +1,130 @@
...
+```bash
+uv pip install -e '.[cortex]'
+uv pip install -r recipes/sft/standalone/requirements.txt
+```
+
+Copy the connection template and fill in your Snowflake host and PAT. To keep the
+PAT out of the file, drop the `pat` key and export `CORTEX_PAT` instead.
+
+```bash
+cp recipes/config.json{.template,}
```
sfc-gh-sbekman: same as reported earlier

sfc-gh-sbekman (changes requested):

recipes/rl/standalone/README.md
sfc-gh-mwyatt: What is borked about this? Would you just prefer `cp recipes/config.json.template recipes/config.json`?

```shell
❯ touch config.json.template
❯ ls config*
config.json.template
❯ cp config.json{.template,}
❯ ls config*
config.json          config.json.template
```

recipes/rl/standalone/requirements.txt
sfc-gh-mwyatt: Let me see if this is required. Just copying over the recipe from https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/main/cortex-client/recipes

sfc-gh-truwase (approved):

recipes/rl/standalone/README.md
sfc-gh-sbekman: oh, I see, that's too cryptic and uses a side effect of shell - not how a normal code would be written

so yes please to `cp recipes/config.json.template recipes/config.json`

recipes/rl/standalone/requirements.txt
sfc-gh-sbekman: I see, but leaning on a potential fragile suggestion to reproduce more of it isn't the best argumentation, especially since you were part of that merge it seems.

it's your call, if you feel it doesn't matter or should be dealt with in the future I've no objection. I was only flagging a potential issue.

sfc-gh-sbekman (approved):

PR #95, opened 2026-08-27
cortex: route forward (no-grad log-probs) through the /operation envelope

sfc-gh-kganesan:
> **Status:** no longer a draft. `forward` has now been exercised against a
> live Cortex zone (see *Live verification* below), which resolves the
> control-plane dependency this was waiting on.
>
> **Review surface:** 3 files, +406 −17. `arctic_platform/client/transports/cortex.py`
> is the change; `tests/client/test_cortex_forward.py` is new and covers it;
> the `pyproject.toml` hunk just makes the transport importable under pytest.
>
> **#110 is stacked on this branch** and blocked on it landing.

## Why

`forward` is in the client's canonical op vocabulary and both on-prem transports implement it, but `CortexTransport` rejected it outright (`NotImplementedError: cortex has no forward`). Any recipe computing log-probs against Cortex dies there — before the control plane is ever contacted. `dss_client.fwd_no_grad()` doesn't help: it is an alias for `forward()`, which posts a base64-JSON `/operation` envelope at a backend route that reads raw DSSST1.

The op itself is not new. On-prem `POST /forward` already splits a DSSST1 batch across DP workers, calls `forward_no_grad`, and returns the merged tensor batch. This PR gives Cortex the same wire; it does not invent semantics.

## What this does

Makes `forward` a first-class octet op alongside `forward-backward`:

- posts a DSSST1 frame as `application/octet-stream`, instead of falling through to `NotImplementedError`
- carries the caller's sub-job token as `?target_sub_job_id=` on the url
- keeps chunking size-driven rather than forced, so a batch under the cap posts as raw DSSST1
- adds `forward` to the group-restart set: replaying it is safe because it mutates no engine state
- result decoding needed no change — tensors survive, since only `generate` is lowered to Python lists

## Corrections in the third commit

The first commit guessed the wire from the client spec and a local `dss-platform` checkout. That checkout was stale; reading the `dss-backend` submodule at the commit Cortex actually pins corrected two things.

**The path is `/forward`.** The zone serves `POST /jobs/{id}/forward` and the ZMD passes the op suffix through verbatim, so the canonical op name already lines up and the per-transport mapping I'd added is gone. Neither `fwd-no-grad` nor `forward-no-grad` exists on that path — the former is only an alias in `gateway_server.py`, a standalone dev gateway the Cortex path doesn't go through.

**The result needs unwrapping.** The zone returns `wire.dumps({job_id, logprobs})` base64'd into a JSON envelope, but *without* the `wire_format` marker `_decode_result` keys off. So the generic path passed the envelope straight through and the caller got a base64 string where tensors belong. `forward` now unwraps it, alongside `generate`'s existing lowering — both moved behind one per-op hook rather than duplicating the branch across `call` and `acall`.

Neither would have surfaced in tests that stub `_send`, which is why both are now pinned: one asserts the outbound path, one asserts the envelope shape the zone actually sends.

Also worth knowing: the Cortex payload is `{job_id, logprobs}` where on-prem's `/forward` returns `{job_id, batch, metrics, avg_loss}`. That divergence is the zone's, and this PR leaves it alone rather than papering over it client-side.

## Why routing has to exist at all

`fwd_no_grad_request()` already sends `reference_model=True` to the `log_prob` sub-job and `False` to `training`. Drop that hint and the request quietly scores against the current policy where reference log-probs were asked for — a plausible-looking wrong number rather than a failure, which is why there is a test pinning it.

The hint still travels even though Cortex answers a log_prob target with `Unimplemented` today: the zone's `/forward` asserts a training zone, and the log_prob zone's `/log-prob` is a separate text-in API, not a tensor forward. An explicit refusal is the outcome we want there. A session can also hold more than one training zone, which the token disambiguates.

## Server side

[snowflake-eng/cortex#156713](https://github.com/snowflake-eng/cortex/pull/156713) is the matching half: the `Forward` control-plane route plus the SnowAPI translation hop. The zone has served `POST /jobs/{id}/forward` all along and the ZMD already proxies op suffixes verbatim, so the control-plane route was the only missing link in the chain.

## Why this was a draft (resolved — see *Live verification*)

Until that control-plane route lands, `forward` reaches a 404 rather than `NotImplementedError` — and 404 is in `_TRANSIENT_STATUSES`, so it retries before surfacing. That is strictly worse to debug than today's behavior. This should merge with or after the server-side change, not before it.

## Test plan

- `tests/client/test_cortex_forward.py` (new, 11 tests): octet body and DSSST1 round-trip, endpoint path, bare frame under the cap, oversized frame splitting under the `forward` label, sub-job routing for both current-policy and reference, `forward-backward` url unchanged, tensors surviving result decode, `log-probs` still raising.
- `pytest tests/client tests/sft` → 147 passed, 5 skipped.
- Two result-decode tests: the zone's real envelope shape (`payload_b64` with no `wire_format`), and a plain-JSON result passing through untouched so an error body isn't mistaken for an envelope.
- **Since verified live:** a full 935-step GSM8K epoch against a live Cortex job on QA6 (see *Live verification*). The original caveat was that the server-side contract was read off pinned sources rather than exercised.

## Follow-ups, deliberately not here

- `cortex-client/dss_client/neutrino_client.py` (same repo, excluded from pre-commit) still base64-wraps byte payloads in `forward()`, and its `fwd_no_grad()` is an alias that adds no no-grad semantics. Recipes route through `CortexTransport`, not this client, so it is a separate change rather than a blocker — but it should follow, so the two clients stop disagreeing about the forward wire.
- `log-probs` (the text-in op) remains unimplemented for Cortex. `forward` is the tensor path.

## Second commit

`[testing]` never pulled `[cortex]`, so `tenacity` was absent under CI's `.[testing,rl]` install and importing `CortexTransport` failed collection. That is why this transport had no tests at all before now. `[testing]` already pulls `[onprem]` for the same reason.

## Live verification

The "not verified against a live GS or zone" caveat below no longer holds. This
path has since carried a full 935-step GSM8K training epoch against a live
Cortex job on QA6, via #110, with `forward` serving Tinker's no-grad log-prob
verb throughout. So the control-plane route this PR was waiting on is present
there, and the wire read off the pinned sources is the wire the zone actually
speaks.

One shape note worth recording against this PR: `forward` returns log-probs as
a **top-level tensor**, while `forward-backward` returns them as nested lists
under `post_process_outputs`. Both are rectangular and padded to full width, so
only the lookup differs — but the published Cortex API spec is stale on the
`post_process_outputs` shape. #110 pins both live shapes in a test so a future
spec change fails a test rather than a training run.


sfc-gh-kganesan:
Closing — this is built against the dedicated octet route in [cortex#156713](https://github.com/snowflake-eng/cortex/pull/156713), which is now closed as superseded. See that PR for the qa6 evidence: forward already reaches the GPU actor through the existing `/operation` dispatch, and the real gap was that nothing decoded the `payload_b64` envelope.

Arctic's transport does still need forward support — the recipes go through this `requests`-based client, not `dss_client`, so neither of the fixes below reaches them on its own:

- [dss-platform#405](https://github.com/snowflake-eng/dss-platform/pull/405) — zone side
- [dss-client#82](https://github.com/snowflake-eng/dss-client/pull/82) — `dss_client` side

The rework targets `/operation` rather than a dedicated endpoint: build the chunk group with the existing `_octet_chunks`, base64 each chunk into the operation envelope, and confirm each leading chunk landed before sending the last one. That last part matters because the zone stages chunks by index but 409s a final chunk that arrives while others are still missing, and `/operation` is asynchronous. The `_NO_CHUNK_OPS` guard added here can go once #405 lands — its error message describes exactly the gap that PR closes.

Branch is preserved; the reusable pieces will come across.

sfc-gh-kganesan:
Reworked the outbound path and verified the whole thing end to end on a live Cortex job (qa6, `dsa-test`, Qwen3-8B on 2xH200, pinned to the debug image carrying dss-platform#410 and dss-client#83).

### What changed

The earlier revision posted a DSSST1 octet body to `/{job}/forward`. That route exists on the zone but not through the control plane, so the op never reached a GPU. `forward` now rides the generic `/operation` RPC:

- the frame travels base64'd in the envelope payload, and the sub-job token moves from `?target_sub_job_id` to a body field;
- chunking happens inside the envelope. The budget applies to the JSON body and base64 inflates the frame 4/3, so the frame slice is the smaller 3/4 share of what the envelope leaves. Constants mirror `dss_client` so both clients split at the same boundary;
- the oversized-forward refusal is retired, since the zone now stages request chunks for `/forward`;
- only the final chunk starts execution, so a last response without a `request_id` now fails loudly rather than leaving us polling a request that never runs.

### End-to-end verification

Ran a client-side loss loop with every hop going through this transport — `fwd_no_grad` over `/operation`, `fwd_bwd` over the octet chunk group, and `step`:

1. fetch log-probs with `fwd_no_grad`;
2. evaluate **TRL's GRPO loss client-side** (`arctic_platform.integrations.trl.loss.trl_grpo` from #84) and take `w = dL/dlogprobs` locally;
3. ship `w` back as a `grpo` surrogate — pin `old_log_probs_shifted` to the log-probs just returned so the ratio is exactly 1, and set `advantages = -w * batch_num_tokens`, at which point `d(grpo)/d(logprobs) == w`;
4. `step`.

The check is self-verifying: on identical weights (steps use `lr=1e-12`, and `grad_norm` is measured pre-update), the surrogate's gradient norm must match a direct server-side `grpo` call carrying the real advantages.

```
direct=99.02234650  surrogate=99.02126312  rel.diff=1.094e-05
PASS: TRL loss client-side, Arctic-Platform transport, Cortex backend
```

The residual is bf16 forward nondeterminism between the two passes. Corroborating server metrics: the direct call reports `approx_kl=0.0500` and `importance_weight=1.0513` (= e^0.05, matching the deliberate offset), while the surrogate reports `approx_kl=6.8e-08` and `importance_weight=1.0000001`, confirming the ratio really is pinned to 1. The client-side loss value also matches the server's own `avg_loss` to 1e-7.

Two things worth recording that this shook out:

- **the RL path is reachable over Cortex.** `_unpickle_batch` accepts `{args, kwargs, context, processing}` and keeps `processing` aside for the worker's RL branch, so `loss_fn` and the post-processing pipeline work over SnowAPI. The `TODO(unify-backends)` note in `client.py` suggesting Cortex only takes an RPC-style batch is stale for this path.
- **client-side loss needs no new server loss function and no image rebuild.** `grpo` at ratio 1 is already a general weighted-log-prob surrogate. I also proved the equivalence offline against the image's real `_grpo_loss`, where parameter gradients agree to 3.3e-16 with and without masking.

Batches here are sent in Cortex's `{"kwargs": ..., "processing": ...}` shape; the transport passes the caller's dict through untouched, which is the contract today. Translating the on-prem `{batch, meta, processing}` shape is still open and deliberately out of scope for this PR.

sfc-gh-kganesan:
Heads-up that #110 is stacked on this branch and is blocked on it landing — retargeting to `main` once this merges.

It uses the no-grad forward path from here to back Tinker's `forward` verb, and exercising it end to end turned up one thing worth recording against this PR: `forward` returns per-token log-probs as a **top-level tensor**, while `forward-backward` returns them as nested lists under `post_process_outputs`. Both are rectangular and padded to full width, so only the lookup differs, but the published Cortex API spec is stale on the `post_process_outputs` shape. #110 pins both live shapes in `TestCortexResponseShapes` so a future spec change fails a test rather than a training run.

The path has now carried a full 935-step GSM8K epoch against a live Cortex job with `kl_sample_train_v1` holding at ~0.001, which is a reasonable soak for the forward routing in this PR.

PR #100, opened 2026-08-31
trl: encode the client-side loss through stock grpo (no server-registered surrogate)

sfc-gh-kganesan:
Builds on #84. Lets the TRL client compute its loss locally and still train over Cortex, without registering a new server-side loss function or rebuilding the image.

## The idea

The client differentiates its own objective to get `w = dL/dlogprobs`, then ships that through the **stock** `grpo` loss by setting `old_log_probs_shifted = logprobs` and `advantages = -w` with `batch_num_tokens=1`. The ratio is then identically 1, and grpo's gradient with respect to the log-probs collapses to exactly `w`. The server never learns which loss the client used.

## This does not replace `weighted_logprob_sum`

`client_loss_encoding` defaults to `"weighted_logprob_sum"`, so #84's path is unchanged and `"grpo"` is opt-in. This is a fallback for servers that do not register the direct loss, not an alternative to it.

Where both are available, prefer the direct encoding: it sends one `[batch, seq]` tensor rather than three, reports a loss value that moves instead of the constant `sum(w)`, and does not lean on incidental properties of a clipped objective that a server-side change could alter silently.

Today the fallback is needed only on Cortex, whose registry is `arctic_training.arctic_rl.processors` in ArcticTraining-dss — not this repo — and holds `causal_cross_entropy`, `grpo` and `grpo_echo_v1` only.

An earlier version of this description said the alternative "needs a Jenkins image rebuild" and framed it as waiting on Cortex to depend on Arctic-Platform. Both were wrong, and read as if this competed with #84. Registering `weighted_logprob_sum` in ArcticTraining is a small addition to a registry that already exists, and doing so retires this encoding. That is the better fix; treat this as a stopgap with a short shelf life.

## What is verified

**The gradient is right, on a real cluster.** Against Qwen3-8B on QA6, the surrogate reproduces the gradient of a direct server-side `grpo` call to a relative difference below `1e-3`, measured through the returned `grad_norm`.

**The tensors line up.** This was the risk I cared most about, because a shift or transpose in the round trip is silent — no error, just a worse model — and a gradient check at a point cannot catch an error consistent across both paths. `weights` round-trips through TRL's packed layout via `_unshift_from_trl` and `_unpack_to_padded`, while `old_log_probs_shifted` is the server's reply untouched, so the two frames have independent provenance.

`TestPayloadAlignment` drives the real `forward_backward` against a mock client with a loss whose derivative is the identity, so the weights coming back out are the log-probs themselves. Every supervised position satisfies `advantages == -old_log_probs_shifted` to `0.0` error, and both encodings weight *exactly* the same positions — grpo does not silently supervise more or fewer tokens than the default.

**It trains, and our gradient is what drives it.** Overfitting a single fixed batch through the real path — client-side NLL, #95's `CortexTransport`, this PR's `_surrogate_payload`, a real optimizer step, and a fresh forward re-measuring each step:

| | start | end | |
|---|---|---|---|
| descent | 6.301 | 5.6e-05 | down on 19/19 steps |
| control, advantages negated | 1.3e-05 | 29.3 | up, after Adam momentum reverses at step ~8 |

client-side loss convergence over Cortex

The control separates "our gradient drives this" from "something moved". Flipping the one sign the encoding depends on reverses the direction, and the turn lands where β₁=0.9 predicts.

<details>
<summary>Per-step records from both runs</summary>

```
DESCENT  (job 74806ebc-2ac5-46fc-8a7f-f23dd5f6bb5c, lr=1e-05, 27 supervised tokens)

step    client NLL   server avg_loss   mean logprob
   0      6.301409      -0.037037037      -0.664602
   1      4.812599      -0.037037037      -0.507579
   2      2.759532      -0.037037037      -0.291044
   3      2.140888      -0.037037037      -0.225797
   4      0.655851      -0.037037037      -0.069172
   5      0.407752      -0.037037037      -0.043005
   6      0.333765      -0.037037037      -0.035202
   7      0.224964      -0.037037037      -0.023727
   8      0.186377      -0.037037037      -0.019657
   9      0.156420      -0.037037037      -0.016497
  10      0.100299      -0.037037037      -0.010578
  11      0.058551      -0.037037037      -0.006175
  12      0.029190      -0.037037037      -0.003079
  13      0.004112      -0.037037037      -0.000434
  14      0.001076      -0.037037037      -0.000113
  15      0.000418      -0.037037037      -0.000044
  16      0.000201      -0.037037037      -0.000021
  17      0.000127      -0.037037037      -0.000013
  18      0.000083      -0.037037037      -0.000009
  19      0.000056      -0.037037037      -0.000006

CONTROL - advantages negated  (job 74806ebc-2ac5-46fc-8a7f-f23dd5f6bb5c, lr=1e-05, 27 supervised tokens)

step    client NLL   server avg_loss   mean logprob
   0      0.000013       0.037037037      -0.000001
   1      0.000013       0.037037037      -0.000001
   2      0.000013       0.037037037      -0.000001
   3      0.000009       0.037037037      -0.000001
   4      0.000010       0.037037037      -0.000001
   5      0.000009       0.037037037      -0.000001
   6      0.000009       0.037037037      -0.000001
   7      0.000009       0.037037037      -0.000001
   8      0.000014       0.037037037      -0.000001
   9      0.000015       0.037037037      -0.000002
  10      0.000018       0.037037037      -0.000002
  11      0.000033       0.037037037      -0.000004
  12      0.000044       0.037037037      -0.000005
  13      0.000191       0.037037037      -0.000020
  14      0.042644       0.037037037      -0.004498
  15      1.931545       0.037037037      -0.203718
  16      3.267599       0.037037037      -0.344630
  17      3.317190       0.037037037      -0.349860
  18      6.730339       0.037037037      -0.709840
  19      7.638395       0.037037037      -0.805612
  20     10.684365       0.037037037      -1.126867
  21     14.444805       0.037037037      -1.523476
  22     20.465557       0.037037037      -2.158477
  23     22.777063       0.037037037      -2.402268
  24     29.332265       0.037037037      -3.093637
```

`server avg_loss` holding constant is expected: `old_log_probs` is refreshed to the current policy each step, so the ratio is identically 1 and the surrogate's value depends only on the advantages. It is the gradient that does the work, and drift in that constant would have meant the encoding was not being read as intended.

</details>

**It runs inside the real TRL trainer, against Cortex.** `send_backward` had never executed with real TRL tensors; it now has, for 150 optimizer steps across two runs of the gsm8k recipe (`recipes/rl/trl/gsm8k`) with the backend swapped from on-prem to Cortex. Real vLLM rollouts, real `accuracy_reward`, real weight sync.

gsm8k over Cortex

The load-bearing number is the right panel. The importance ratio stayed within **0.9993–1.0025** across all 120 steps, which is an end-to-end check on the log-prob frame through packing, unpacking and shifting — an off-by-one anywhere would push it far from 1.

The reward went from 0.497 (first 30 steps) to 0.633 (last 30), `t ≈ 2.3`. I would call that suggestive, not a convergence result: the steps are not independent samples (32 prompts recycled for 120 steps, autocorrelated), and 13% of steps had `reward_std = 0`, meaning every rollout in the group scored alike and the step contributed no gradient at all. A real claim needs the recipe's scale — 7,473 prompts on 8 GPUs, not 32 prompts on 1.

**The two PRs compose.** #95 and this branch merge with no conflicts, and the union of both test suites is green (77 passed).

## The blocker this surfaced

Getting the recipe to Cortex at all required an adapter, for reasons independent of this PR. Every call in `arctic_platform/integrations/trl/client.py` sends the on-prem `{batch, meta, processing}` envelope and asks for the post-processors `apply_temperature` and `compute_entropy_and_logprobs`. Grep `dss/` for those two and you get nothing — a Cortex zone registers only `compute_logprobs`. The first failure is the zone rejecting the envelope outright: `dispatch requires 2D input_ids`.

The [adapter](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/pr100/gsm8k/cortex_trl_adapter.py) translates the envelope and substitutes `compute_logprobs`, which is only sound because the recipe runs at `temperature=1.0` where `apply_temperature` is a no-op; anything else is refused rather than silently mis-scaled. Entropy comes back as zeros, so any entropy metric is meaningless and an entropy bonus would silently do nothing.

That is a shim for a demonstration, not a fix. The real fix is either teaching Cortex zones those post-processors or teaching the integration to speak both dialects, and it probably wants its own issue.

Two smaller ones, both worth knowing: the Cortex image ships FA3 but not FA2 (the recipe asks for FA2 and the job dies at model load), and `offload_optimizer: {"device": "none"}` is still enough for DeepSpeed to instantiate CPUAdam, which then asserts because the params are on GPU.

## Not verified

- **Convergence at the recipe's scale.** See above — 1 GPU and 32 prompts is not 8 GPUs and 7,473.
- **The on-prem backend.** The encoding has only ever run against Cortex.
- **`temperature != 1.0`**, and entropy anywhere.
- **`loss_mask`.** `_surrogate_payload` derives it from `attention_mask`. When a batch also carries `labels` that is one position too generous per row and grpo's preflight rejects it, so both harnesses override it with `labels != -100`. TRL's own batches carry no `labels`, so the default is correct on the path this PR targets.

## Note on the equivalence

The `grpo` encoding and `weighted_logprob_sum` produce **identical gradients**, not identical loss values — the reported scalars differ (`sum(w)` versus `sum(w * logprobs)`). An earlier draft of this description overstated that as full equivalence. `test_client_loss_encoding.py` pins both facts.

Raw artifacts, the gsm8k runner and the adapter live on [`karthik/pr100-assets`](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/pr100), kept off this branch so the diff stays code-only. Safe to delete once this merges.

### DIFF
--- a/arctic_platform/integrations/trl/__init__.py
+++ b/arctic_platform/integrations/trl/__init__.py
@@ -17,10 +17,9 @@
 
 from typing import TYPE_CHECKING
 
-from arctic_platform.integrations.trl.client import ArcticOptimizer
-from arctic_platform.integrations.trl.client import ArcticTrainingClient
-
 if TYPE_CHECKING:
+    from arctic_platform.integrations.trl.client import ArcticOptimizer
+    from arctic_platform.integrations.trl.client import ArcticTrainingClient
     from arctic_platform.integrations.trl.rollout import ArcticRolloutWorker
     from arctic_platform.integrations.trl.weights import ArcticWeightTransfer
     from arctic_platform.rl.processors.weighted_logprob import weighted_logprob_sum
@@ -35,7 +34,13 @@ __all__ = [
 
 
 def __getattr__(name: str):
-    # Lazy: rollout/weights need async_grpo; loss needs the server.
+    # Lazy: client needs trl, rollout/weights need async_grpo; loss needs the server.
+    # Importing client eagerly made the whole package -- loss included -- unimportable
+    # without trl, which silently skipped the CPU loss tests.
+    if name in ("ArcticOptimizer", "ArcticTrainingClient"):
+        from arctic_platform.integrations.trl import client
+
+        return getattr(client, name)
     if name == "ArcticRolloutWorker":
         from arctic_platform.integrations.trl.rollout import ArcticRolloutWorker
 
--- a/arctic_platform/integrations/trl/client.py
+++ b/arctic_platform/integrations/trl/client.py
@@ -24,6 +24,9 @@ from typing import Any
 import torch
 from trl.experimental.api import ForwardBackwardOutput
 
+from arctic_platform.integrations.trl.loss import _CLIENT_LOSS_ENCODINGS
+from arctic_platform.integrations.trl.loss import _surrogate_payload
+
 
 def _meta_dict(
     *,
@@ -67,6 +70,11 @@ class ArcticTrainingClient:
             (perf + parity with verl/SkyRL). When ``False`` (default), keep the
             two-pass client-side surrogate path.
         server_loss_fn: Server GRPO loss name/dotted-path used when ``server_side_loss``.
+        client_loss_encoding: How the two-pass path expresses ``dL/dlogprobs`` to the
+            server. ``"weighted_logprob_sum"`` (default) names this package's own
+            surrogate, which the server must have registered. ``"grpo"`` encodes the
+            same quantity in the stock GRPO loss instead, so the path runs on a server
+            that ships nothing from this package -- see :func:`_surrogate_payload`.
     """
 
     def __init__(
@@ -83,6 +91,7 @@ class ArcticTrainingClient:
         logits_compute_in_fp32: bool = False,
         server_side_loss: bool = False,
         server_loss_fn: str = "arctic_platform.integrations.trl.loss.trl_grpo",
+        client_loss_encoding: str = "weighted_logprob_sum",
         zorro_train_enable: bool = False,
         response_len: int | None = None,
         zorro_load_balancer: bool = False,
@@ -99,6 +108,11 @@ class ArcticTrainingClient:
         self.logits_compute_in_fp32 = logits_compute_in_fp32
         self.server_side_loss = server_side_loss
         self.server_loss_fn = server_loss_fn
+        if client_loss_encoding not in _CLIENT_LOSS_ENCODINGS:
+            raise ValueError(
+                f"client_loss_encoding must be one of {sorted(_CLIENT_LOSS_ENCODINGS)}, got {client_loss_encoding!r}"
+            )
+        self.client_loss_encoding = client_loss_encoding
         self.zorro_train_enable = zorro_train_enable
         # Padded response width (== configured max_completion_length == server ds_worker_config.response_len).
         # Zorro emits verl-style structured [B, max_prompt_len + response_len] batches, so this must be set.
@@ -188,14 +202,20 @@ class ArcticTrainingClient:
         def send_backward(grad_loss: torch.Tensor) -> None:
             # Scale, unshift to [1, T], unpack to [B, S].
             weights = _unpack_to_padded(_unshift_from_trl(grad_log_probs * grad_loss), seq_lens)
+            back_batch, back_loss_fn, loss_config = _surrogate_payload(
+                self.client_loss_encoding, batch, weights, out["logprobs"], self.loss_fn
+            )
+            processing = {
+                "post": ["apply_temperature", "compute_entropy_and_logprobs"],
+                "loss_fn": back_loss_fn,
+            }
+            if loss_config:
+                processing["config"] = loss_config
             self.client.fwd_bwd(
                 {
-                    "batch": {**batch, "logprob_weights_shifted": weights},
+                    "batch": back_batch,
                     "meta": self._meta(calculate_entropy=False),
-                    "processing": {
-                        "post": ["apply_temperature", "compute_entropy_and_logprobs"],
-                        "loss_fn": self.loss_fn,
-                    },
+                    "processing": processing,
                 }
             )
 
--- a/arctic_platform/integrations/trl/loss.py
+++ b/arctic_platform/integrations/trl/loss.py
@@ -65,3 +65,48 @@ def trl_grpo(
     loss = masked / batch_num_tokens * dp_size / grad_accum
 
     return loss, {"loss": loss.detach()}
+
+
+_CLIENT_LOSS_ENCODINGS = ("weighted_logprob_sum", "grpo")
+
+
+def _surrogate_payload(
+    encoding: str,
+    batch: dict,
+    weights: torch.Tensor,
+    old_log_probs: torch.Tensor,
+    loss_fn: str,
+) -> tuple[dict, str, dict]:
+    """Express ``w = dL/dlogprobs`` as a batch the server can differentiate.
+
+    Both encodings ask for the same thing -- a loss whose gradient wrt log-probs is
+    exactly ``w`` -- and differ only in which registered loss carries it.
+
+    ``weighted_logprob_sum`` states it directly, but the server has to have this
+    package installed. ``grpo`` gets there with a loss every RL server already
+    ships: its gradient wrt log-probs is ``-advantages * ratio``, so pinning
+    ``old_log_probs_shifted`` to the log-probs the forward pass just returned makes
+    the ratio 1 and leaves ``-advantages``.
+
+    Scale follows from ``token-mean``, which is ``masked_sum / batch_num_tokens *
+    dp_size``. Setting both to 1 collapses it to a plain masked sum, so ``grpo``
+    reduces to exactly what ``weighted_logprob_sum`` computes -- including how it
+    behaves under data parallelism, which is inherited here rather than redefined.
+
+    Clipping cannot engage: at ratio 1, ``min(r*A, clip(r)*A)`` is ``A`` for any
+    epsilon. The server recomputes log-probs in bf16, so the ratio lands at 1 only
+    to within forward nondeterminism, which perturbs the gradient by ~1e-5
+    relative -- far inside the clip band, but not bit-exact.
+    """
+    if encoding == "grpo":
+        return (
+            {
+                **batch,
+                "old_log_probs_shifted": old_log_probs,
+                "advantages": -weights,
+                "loss_mask": batch["attention_mask"].to(weights.dtype),
+            },
+            "grpo",
+            {"batch_num_tokens": 1, "dp_size": 1},
+        )
+    return {**batch, "logprob_weights_shifted": weights}, loss_fn, {}
--- a/arctic_platform/integrations/trl/loss.py
+++ b/arctic_platform/integrations/trl/loss.py
@@ -89,9 +89,16 @@ def _surrogate_payload(
     the ratio 1 and leaves ``-advantages``.
 
     Scale follows from ``token-mean``, which is ``masked_sum / batch_num_tokens *
-    dp_size``. Setting both to 1 collapses it to a plain masked sum, so ``grpo``
-    reduces to exactly what ``weighted_logprob_sum`` computes -- including how it
-    behaves under data parallelism, which is inherited here rather than redefined.
+    dp_size``. Setting both to 1 collapses it to a plain masked sum, leaving the
+    gradient unnormalized exactly as ``weighted_logprob_sum`` leaves it -- so the
+    two encodings behave identically under data parallelism, inheriting that
+    behaviour rather than redefining it.
+
+    What matches is the gradient, not the number. ``weighted_logprob_sum`` reports
+    ``sum(w * logprobs)`` while this reports ``sum(w)``, because at ratio 1 the
+    per-token GRPO term is ``-advantages``. Both differentiate to ``w``, which is
+    all the backward pass consumes; the client reports its own loss to TRL and
+    discards the server's, so the difference is confined to server-side metrics.
 
     Clipping cannot engage: at ratio 1, ``min(r*A, clip(r)*A)`` is ``A`` for any
     epsilon. The server recomputes log-probs in bf16, so the ratio lands at 1 only
--- a/tests/integrations/trl/test_client_loss_encoding.py
+++ b/tests/integrations/trl/test_client_loss_encoding.py
@@ -32,6 +32,7 @@ import torch
 
 from arctic_platform.integrations.trl.loss import _surrogate_payload
 from arctic_platform.rl.processors.grpo import grpo_loss
+from arctic_platform.rl.processors.weighted_logprob import weighted_logprob_sum
 
 B, S = 2, 6
 
@@ -110,6 +111,49 @@ class TestGrpoEncoding:
         payload, _, config = _surrogate_payload("grpo", batch, weights, log_probs, "unused")
 
         assert torch.allclose(_server_gradient(payload, config, log_probs), weights, atol=1e-12)
+
+
+class TestEquivalenceWithTheDefaultEncoding:
+    def test_both_encodings_produce_the_same_gradient(self):
+        """The two paths must be interchangeable in the only way that matters."""
+        torch.manual_seed(2)
+        batch = _batch()
+        log_probs = torch.randn(B, S, dtype=torch.float64)
+        weights = torch.randn(B, S, dtype=torch.float64)
+        weights[:, -1] = 0.0  # as _unpack_to_padded leaves padded positions
+
+        grpo_payload, _, config = _surrogate_payload("grpo", batch, weights, log_probs, "x")
+        wls_payload, _, _ = _surrogate_payload("weighted_logprob_sum", batch, weights, log_probs, "x")
+
+        leaf = log_probs.clone().requires_grad_(True)
+        wls_loss, _ = weighted_logprob_sum({"logprobs": leaf}, wls_payload, {}, {}, "cpu")
+        (wls_grad,) = torch.autograd.grad(wls_loss, leaf)
+
+        assert torch.allclose(_server_gradient(grpo_payload, config, log_probs), wls_grad, atol=1e-12)
+
+    def test_the_reported_loss_value_differs_and_that_is_fine(self):
+        """Only the gradient is shared; the scalars are not the same number.
+
+        At ratio 1 the per-token GRPO term is ``-advantages``, so grpo reports
+        ``sum(w)`` where the default reports ``sum(w * logprobs)``. The client
+        reports its own loss to TRL and drops the server's, so this shows up only
+        in server-side metrics -- but it would be a real surprise when comparing
+        runs across encodings, so it is pinned rather than left implicit.
+        """
+        torch.manual_seed(3)
+        batch = _batch()
+        log_probs = torch.randn(B, S, dtype=torch.float64)
+        weights = torch.randn(B, S, dtype=torch.float64)
+
+        grpo_payload, _, config = _surrogate_payload("grpo", batch, weights, log_probs, "x")
+        wls_payload, _, _ = _surrogate_payload("weighted_logprob_sum", batch, weights, log_probs, "x")
+
+        grpo_value, _ = grpo_loss({"logprobs": log_probs}, grpo_payload, config, "cpu")
+        wls_value, _ = weighted_logprob_sum({"logprobs": log_probs}, wls_payload, {}, {}, "cpu")
+
+        assert float(grpo_value) == pytest.approx(float(weights.sum()))
+        assert float(wls_value) == pytest.approx(float((weights * log_probs).sum()))
+        assert float(grpo_value) != pytest.approx(float(wls_value))
 
 
 class TestDefaultEncoding:
--- a/tests/integrations/trl/test_client_loss_encoding.py
+++ b/tests/integrations/trl/test_client_loss_encoding.py
@@ -156,6 +156,73 @@ class TestEquivalenceWithTheDefaultEncoding:
         assert float(grpo_value) != pytest.approx(float(wls_value))
 
 
+class TestPayloadAlignment:
+    """Whether the tensors line up once the real client builds them.
+
+    The encoding tests above feed ``_surrogate_payload`` tensors that are aligned
+    by construction. In the client they are not: ``weights`` round-trips through
+    TRL's packed layout via ``_unshift_from_trl`` and ``_unpack_to_padded``, while
+    ``old_log_probs_shifted`` is the server's reply untouched. If those two frames
+    disagree by even one position the gradient is silently wrong -- no error, just
+    a worse model -- so it is worth pinning against the real call path.
+    """
+
+    LENS = (5, 4)
+
+    class _MockClient:
+        def __init__(self, lens):
+            self.payload = None
+            b, s = len(lens), max(lens)
+            # Distinctive per-position values, so a misalignment cannot look plausible.
+            self.logprobs = torch.zeros(b, s, dtype=torch.float32)
+            for row, n in enumerate(lens):
+                for col in range(n):
+                    self.logprobs[row, col] = -(10 * (row + 1) + col)
+
+        def fwd_no_grad(self, payload):
+            del payload
+            return {"batch": {"logprobs": self.logprobs.clone(), "entropy": torch.zeros_like(self.logprobs)}}
+
+        def fwd_bwd(self, payload):
+            self.payload = payload
+            return {}
+
+    def _run(self, encoding: str):
+        client_mod = pytest.importorskip("arctic_platform.integrations.trl.client")
+        mock = self._MockClient(self.LENS)
+        client = client_mod.ArcticTrainingClient(client=mock, client_loss_encoding=encoding)
+        total = sum(self.LENS)
+
+        out = client.forward_backward(
+            model=None,
+            input_ids=torch.arange(1, total + 1).unsqueeze(0),
+            position_ids=torch.cat([torch.arange(n) for n in self.LENS]).unsqueeze(0),
+            completion_mask=torch.ones(1, total, dtype=torch.long),
+            # dL/dlogprobs == logprobs, so the weights coming back out are the
+            # log-probs themselves after the full round trip.
+            loss_fn=lambda leaf: (leaf**2).sum() / 2,
+        )
+        out.loss.backward()
+        return mock
+
+    def test_advantages_are_paired_with_the_logprobs_they_came_from(self):
+        mock = self._run("grpo")
+        batch = mock.payload["batch"]
+
+        assert torch.equal(batch["old_log_probs_shifted"], mock.logprobs)
+        carried = batch["advantages"] != 0
+        assert carried.any()
+        assert torch.allclose(batch["advantages"][carried], -mock.logprobs[carried])
+
+    def test_both_encodings_weight_exactly_the_same_positions(self):
+        """grpo must not silently supervise more or fewer tokens than the default."""
+        grpo_batch = self._run("grpo").payload["batch"]
+        default_batch = self._run("weighted_logprob_sum").payload["batch"]
+
+        assert torch.equal(grpo_batch["advantages"] != 0, default_batch["logprob_weights_shifted"] != 0)
+        assert torch.allclose(grpo_batch["advantages"], -default_batch["logprob_weights_shifted"])
+
+
 class TestDefaultEncoding:
     def test_default_is_unchanged(self):
         """The existing path must keep naming its own loss and its own key."""
--- a/tests/integrations/trl/test_client_loss_encoding.py
+++ b/tests/integrations/trl/test_client_loss_encoding.py
@@ -1,0 +1,134 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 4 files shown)


sfc-gh-kganesan:
@sfc-gh-truwase thanks — you're right on the substance, and the description was the cause. I've rewritten the motivation section. Point by point:

| Your concern | Answer |
|---|---|
| Did you run this against a Cortex that supports `fwd_no_grad`? | Yes, and against the **default deployed** image — no debug tag pinned. The importance ratio held 0.9993–1.0025 across all 120 steps, which is an end-to-end check that the log-prob frame is right. |
| Client-side loss already works in #84 via `weighted_logprob_sum` — does this replace it? | No. `client_loss_encoding` defaults to `"weighted_logprob_sum"`; `"grpo"` is opt-in. Your path is untouched. |
| Do the two encodings agree? | Identical **gradients**, different reported loss values (`sum(w)` vs `sum(w · logprobs)`). Both pinned in `test_client_loss_encoding.py`. |
| Is the "needs a Jenkins image rebuild" motivation valid? | True today, but overstated, and I named the wrong repo. Corrected below. |
| Is this PR unaware Cortex will depend on AP for processors and losses? | Partly fair — but the premise is off in a way that helps us. Cortex sources losses from **ArcticTraining-dss**, not AP. |
| Faster path — guide updates for the new Cortex? | Yes, and the loss was never the blocker. List below. |

## The correction that matters

Cortex's loss registry is `arctic_training.arctic_rl.processors` in ArcticTraining-dss, not Arctic-Platform. dss-platform declares no `arctic_platform` dependency and contains no `weighted_logprob` anywhere. That registry holds exactly `causal_cross_entropy`, `grpo` and `grpo_echo_v1`.

So Cortex doesn't need to wait on depending on AP for this. **Registering `weighted_logprob_sum` in ArcticTraining is a small addition to a registry that already exists, and it retires this PR's encoding entirely.** I'd rather that happened than land a second code path we then delete. If you agree, I'll take it to @thong-nguyen_snow — it's a better fix than either of our PRs.

Worth noting the capability handshake is already an established pattern there: `grpo_echo_v1`'s docstring says requesting a loss name on a server without support "fails at loss-fn resolution, which is the capability handshake." So the client can prefer the direct encoding and fall back only on that failure, which self-corrects the moment Cortex registers it — no flag for the caller to get wrong.

## Related update on your first question

Thong merged `fwd_no_grad` for the Neutrino backend into ArcticTraining's `ArcticRLClient` on Aug 31 ([ArcticTraining-dss#158](https://github.com/snowflake-eng/ArcticTraining-dss/pull/158)), riding the same generic `forward` operation with chunked DSSST1 results.

There are two parallel client stacks here. His is done. AP's own `CortexTransport` still has `_WIRE_OPERATION = {"forward-backward": ..., "generate": ...}` with no `forward` and raises `cortex has no forward` — that's what #95 fixes, and it's the stack `integrations/trl` goes through. So if your pipeline uses AP's client you need #95; if it uses ArcticTraining's you already have it.

## What you actually need to run against real Cortex

The loss encoding isn't the blocker — the envelope dialect is.

1. **#95**, for the transport's `forward` route (only if you're on AP's client).
2. **An envelope adapter.** `integrations/trl/client.py` sends the on-prem `{batch, meta, processing}` shape and requests the post-processors `apply_temperature` and `compute_entropy_and_logprobs`. Grep `dss/` for those two and you get nothing; a Cortex zone registers only `compute_logprobs`. First failure is the zone refusing the request with `dispatch requires 2D input_ids`, before any model call. My [shim](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/pr100/gsm8k/cortex_trl_adapter.py) translates it, but it is only sound at `temperature=1.0` where `apply_temperature` is a no-op, and it returns zeros for entropy — so entropy metrics are meaningless and an entropy bonus would silently do nothing. This wants its own issue; it blocks you regardless of which encoding wins.
3. **FA3, not FA2.** The recipe asks for `flash_attention_2` and the Cortex image ships only FA3, so the job dies at model load.
4. **Drop `offload_optimizer` entirely.** `{"device": "none"}` is still enough for DeepSpeed to instantiate CPUAdam, which then asserts because the params are on GPU.

## What I'd keep from this PR either way

`TestPayloadAlignment` drives the real `forward_backward` against a mock client whose loss has an identity derivative, so the weights coming back out are the log-probs themselves, and every supervised position satisfies `advantages == -old_log_probs_shifted` to `0.0`. That exercises `_unshift_from_trl` and `_unpack_to_padded` — shared code that produces `weights` for **your** path too, and `send_backward` had never run against real TRL tensors before. Worth keeping even if the grpo encoding goes away.

sfc-gh-kganesan:
## Convergence check on the gsm8k recipe, three seeds, with a falsification control

@sfc-gh-truwase I re-ran correctness the way you validated #84 — your `qwen3_1.7b_gsm8k_grpo` config, your seeds (7/42/123), 100 steps, Qwen3-1.7B — but over Cortex with the client-side loss going through this PR's `grpo` encoding.

**One deviation, and I want to flag it rather than bury it.** I could not reproduce your baseline arm: this box has no GPUs (no `nvidia-smi`), and your baseline is native TRL with a local `vllm serve`, which can't run on Cortex since it needs a colocated sampler. So instead of *native TRL vs Arctic*, I ran *correct sign vs flipped sign* — every seed twice, identical config, with only `advantages = -weights` negated. That controls for the thing I actually care about: whether reward moves because of the gradient we encode, or for some unrelated reason (sampler warmup, dataset order, reward-function drift).

### Results

| seed | arm | first 30 | last 30 | delta | slope /100 | ratio | kl |
|---|---|---|---|---|---|---|---|
| 7 | client-side loss | 0.4624 | 0.6539 | **+0.191** | +0.288 | 1.0006 | 6.6e-04 |
| 7 | sign flipped | 0.1398 | 0.0010 | −0.139 | −0.207 | 1.0007 | 7.1e-04 |
| 42 | client-side loss | 0.4675 | 0.6183 | **+0.151** | +0.260 | 1.0006 | 6.7e-04 |
| 42 | sign flipped | 0.1612 | 0.0020 | −0.159 | −0.242 | 1.0007 | 7.5e-04 |
| 123 | client-side loss | 0.4749 | 0.6275 | **+0.153** | +0.252 | 1.0006 | 6.5e-04 |
| 123 | sign flipped | 0.1390 | 0.0000 | −0.139 | −0.209 | 1.0007 | 8.2e-04 |

Across the three seeds: **+0.165 ± 0.013** (sem) for the real arm, **−0.146 ± 0.007** for the control, a paired separation of **+0.311 ± 0.011**. The separation is the same size on all three seeds, and the between-seed spread (sd 0.019) is 16x smaller than the effect, so this isn't seed luck. `ratio` sits at 1.0006 and `kl` at ~7e-04 on every run, which is the same no-drift condition you used on #84.

three seeds

The control is the informative half of that picture: flipping one sign drives accuracy reward to *exactly* 0.0 by step ~30 and holds it there. Both arms start in the same range (0.26–0.32 at step 1) and then go opposite directions.

### What this does and does not establish

It establishes that training over Cortex under this encoding optimises the intended objective, at recipe scale, reproducibly, and that the improvement is causally attributable to the gradient we send. Combined with the exact checks already in the PR body — identical gradients for both encodings on CPU, agreement with a direct server-side `grpo` call to <1e-3, and `advantages == -old_log_probs_shifted` to 0.0 at every supervised position — I think the encoding is correct.

It does **not** establish parity with native TRL in absolute reward, because I can't run that arm. If you want that comparison, the encoding A/B is a one-line change on your setup and is the stronger test anyway: run your recipe with `client_loss_encoding="weighted_logprob_sum"` vs `"grpo"` at one seed. Both encodings are available on-prem, they should produce identical gradients, and if they do, this PR is a pure no-op on your path.

Repro: `run_seed_matrix.sh` + `analyze_seed_matrix.py`. One caveat on my side — with `num_prompts=512` and up to 8 groups/step, the tail of a 100-step run can revisit prompts, so treat the absolute reward level as indicative and the paired separation as the result.

sfc-gh-kganesan:
Harness, the six per-run metrics files, and the analysis output are on [`karthik/pr100-assets`](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/harness) so the numbers above are auditable without a rerun. Nothing there is part of this PR's diff.

sfc-gh-kganesan:
## Correction on the image question, and it lands in your favour

@sfc-gh-truwase two corrections to my earlier comments, then what changed today.

**1. The three-seed convergence run pinned a debug image.** It used `dev_20260828_170144_40ee0a90875`. I did not say so in that comment and I should have, because it was the answer to your first question. (The separate 120-step run in the comment before it really was on the default image — that claim stands.)

**2. I also mis-diagnosed why the pin was needed, and the right diagnosis is a smaller problem.** I said the default image could not decode a *chunked* `/operation` payload. Chunking was never involved:

- AP's transport only splits a forward frame above 12.58 MB (`_OPERATION_MAX_JSON_BYTES` minus envelope overhead, times 3/4 for base64).
- The forward that failed is `engine_old_log_probs` on one rollout group — 8 sequences of at most 1024 tokens, a couple hundred KB. It went as a single envelope, always.
- It failed at `0%| 0/100`, on the very first forward, so payload growth over the run wasn't it either.

The actual error was `_unpickle_batch` → `wire.loads(body)` on a body that wasn't a DSSST1 frame. `header too large` is what safetensors says when it gets base64 text instead of decoded bytes. That's the **envelope decode** half of Thong's Sep 1 work, not the chunking half.

## As of today the pin is unnecessary

Re-ran on QA6 against the **default** deployed image, no debug tag, through AP's own `CortexTransport` — the stack `integrations/trl` uses:

| check | result |
|---|---|
| `fwd_no_grad` via `/operation` | returns logprobs, `(2, 128)` |
| direct server-side `grpo` `fwd_bwd` | `grad_norm` 99.02342224 |
| client-side TRL GRPO → this PR's `grpo` surrogate | `grad_norm` 99.02857971 |
| relative difference | **5.2e-05** |

So the client-side loss path runs end to end on a stock Cortex with nothing pinned. Thong's fix reached the default deployment sometime between Sep 2 and today.

Separately, and only because it matters for larger batches later: I forced a forward into 33 chunks by dropping the client's JSON ceiling to its 16 KB floor, and the assembled result came back **bitwise identical** to the single-envelope forward. Chunk assembly is correct on the default image too, we just never reach it at this recipe's scale.

## Your "is it a no-op on my path" question is now a test

`e68cef1` adds `test_both_encodings_hand_the_server_the_same_gradient`. The equivalence test I had before differentiates the two losses on tensors `_surrogate_payload` was handed, which are aligned by construction — it cannot see a divergence introduced anywhere else in `forward_backward`. The new one runs the real client under both encodings and differentiates the real server losses on whatever each one actually produced.

Verified it can fail: rolling `advantages` by one position breaks it along with five existing tests. So "selecting `grpo` changes nothing the server acts on" is checkable rather than argued, on CPU, no Cortex needed.

## On the opt-in, which I think is your real objection

You asked why the caller should have to know. I agree they shouldn't, and I now have the two facts needed to remove it rather than argue about it.

**The failure is attributable.** Cortex resolves loss names through `arctic_training.arctic_rl.processors._resolve_fn`, which falls back to a dotted-path import — and AP's default loss name *is* a dotted path. So a stock Cortex gives exactly:

```
ModuleNotFoundError: No module named 'arctic_platform.rl.processors.weighted_logprob'
  at pipeline.py:165 resolve_packed_loss_reduction -> _resolve_fn -> importlib.import_module
```

Distinguishable from an OOM or a bad batch, and it names the module, so a fallback can key on it precisely.

**The failure is clean.** `resolve_packed_loss_reduction` runs before the microbatch loop reaches `engine.backward`, so nothing is accumulated. Confirmed on QA6: same request before and after a failed attempt moved `grad_norm` by 2.0e-04, against 4.4e-05 scatter between two identical requests. An automatic retry that double-counted a microbatch would move it by order 1, not 0.02%.

So the client can ask for its own loss, catch that specific error once, and fall back — which self-corrects the moment Cortex registers `weighted_logprob_sum`, with no flag for anyone to set.

One thing that argues against just *always* using `grpo` and deleting the choice: at a realistic 8×512 shape the `grpo` payload is 144 KiB against 112 KiB, because it adds `loss_mask` and `old_log_probs_shifted`. A 29% larger backward request on your on-prem path for no benefit there, so I'd rather auto-select than standardise on one.

**It's your branch, so I'd rather ask than assume:** do you want the auto-fallback in this PR, or should this land as the opt-in it is and the fallback come separately? I'm happy to write it either way — I just don't want to change the default behaviour of your path without you agreeing to it.

Scripts and raw logs for everything above are on `karthik/pr100-assets`: the two new probes and their output under [`harness/`](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/harness) and [`harness/defaultimage/`](https://github.com/Snowflake-AI-Research/Arctic-Platform/tree/karthik/pr100-assets/harness/defaultimage), and the transport check in [`pr100/ap_client_side_loss.py`](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/karthik/pr100-assets/pr100/ap_client_side_loss.py). None of it is in this PR's diff.

sfc-gh-kganesan:
## Pickup instructions, if you want to hand this to an agent

Wrote up exactly what's needed to use this PR for client-side loss on top of #84: **[PR100-FOR-84.md](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/karthik/pr100-assets/PR100-FOR-84.md)** on `karthik/pr100-assets`.

The short version, because the headline is easy to get wrong:

- **On-prem: ignore this PR.** #84 already does client-side loss through `weighted_logprob_sum`, and `d71b6c94` put it somewhere an on-prem server can import. Leave `client_loss_encoding` unset and nothing about your path changes. `test_both_encodings_hand_the_server_the_same_gradient` checks that on CPU if you'd rather not take my word.
- **Cortex: this PR alone is not enough**, and that's the part worth flagging. You need three things, and only the first is in #100:
  1. #100, for `client_loss_encoding="grpo"` — Cortex's registry has no `weighted_logprob` at all, so #84's default loss name fails to import there.
  2. **#95**, for the transport. #100's branch still has `_WIRE_OPERATION` without `forward`, so a no-grad forward raises `cortex has no forward`.
  3. **An envelope adapter, which is in no PR yet.** `integrations/trl` speaks the on-prem dialect and asks for `apply_temperature` and `compute_entropy_and_logprobs`; neither name exists anywhere in ArcticTraining-dss, which registers only `identity` and `compute_logprobs`. Reference impl is on the assets branch, but it's only sound at `temperature=1.0` and returns zeros for entropy. This blocks Cortex whichever loss encoding wins, so it probably deserves its own PR.

The doc also covers the two job-config settings that kill a job at startup (FA3 not FA2; drop `offload_optimizer` entirely rather than setting it to `"none"`), and a one-operation check that verifies the encoding and transport in a couple of minutes so you don't burn a recipe run finding out.

No debug image needed any more — that's the correction in my previous comment.

sfc-gh-kganesan:
## Self-review of the handoff, and a fix that matters if anyone tries to reproduce this

I re-ran my own instructions from a clean checkout instead of re-reading them. Two things worth reporting.

**The published harness was not runnable off my machine.** `cortex_client_side_loss.py` and `ap_client_side_loss.py` both called `sys.path.insert(0, "<my clone>")` unconditionally, which silently beat the caller's `PYTHONPATH`. For scripts whose only purpose is evidence, that's the worst possible bug: you couldn't tell which `dss-client` or which Arctic-Platform checkout a run had actually exercised. Several flags also defaulted to absolute paths in my home directory, and `harness/` shipped probes that imported a module living only in `pr100/`.

Fixed in `571e4b9`: defaults are now derived from the imported packages, so the loss under test always matches the transport under test. Verified `default_trl_repo()` and `default_job_body()` follow `PYTHONPATH` and resolve to real files.

**`run_seed_matrix.sh` was still telling readers to pin the debug image**, and justifying it with the chunk-assembly diagnosis that turned out to be wrong. Marked historical with the actual cause, and a note that a fresh run should just drop the flag.

### What I verified by executing it

| check | result |
|---|---|
| #100 applies to #84's head | 4 ahead, 0 behind, no rebase |
| #100 + #95 merge | clean, no conflicts |
| `_WIRE_OPERATION` after merge | gains `"forward"` |
| tests on #100 alone | `12 passed`, bare `pytest` from repo root, no `PYTHONPATH` |
| tests on the combination | `78 passed` |
| the equivalence test can fail | rolling `advantages` one position breaks it plus 5 others |
| every error string in the doc | located at the cited file and line |

The `dispatch requires 2D input_ids` one, for instance, is `dss/ray_dss/jobs/gpu/sp/data_plane.py:259`.

### Still not verified, stated plainly in the doc

The recipe has **not** been re-run at 100 steps on the default image — the three-seed result was produced on the pinned debug image. The transport check supersedes it for the transport question, but there is no recipe-scale default-image run. The adapter has also never been exercised at any temperature other than 1.0.

Instructions: **[PR100-FOR-84.md](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/karthik/pr100-assets/PR100-FOR-84.md)**. It leads with the on-prem/Cortex split, because on-prem needs nothing from this PR and Cortex needs two things that aren't in it.

sfc-gh-kganesan:
### The adapter is now #109, and a scoping correction on my own evidence

Two things, one of them a correction.

**The correction.** I've been letting yesterday's default-image transport check stand in for "the Cortex path works end to end." It doesn't, and the difference matters for reading the numbers in this PR.

`ap_client_side_loss.py` calls `transport.call(fwd_no_grad_request(jobs, {"kwargs": batch}))`. It writes the *Cortex* envelope by hand and loads only `integrations/trl/loss.py`. It never touches `ArcticTrainingClient`. So it proves the transport route and the surrogate gradient math — the `5.2e-05` agreement is real — but it bypasses precisely the layer that has the dialect mismatch, because it does that layer's job inline.

The only full-path evidence is the three-seed convergence run, and that went through a `CortexTRLAdapter` sitting in my scratch directory. Which means: **no Cortex end-to-end result in this PR was reproducible from the PRs alone.** Anyone starting from #84 + #95 + #100 would have hit `dispatch requires 2D input_ids` at the zone. That's on me for not flagging the dependency sooner.

| layer | covered by | reproducible from PRs before today? |
|---|---|---|
| `CortexTransport.forward` | transport check, default image | yes, with #95 |
| surrogate gradient math | transport check + CPU tests | yes |
| `ArcticTrainingClient` envelope → Cortex | seed runs only | **no** |

**The fix.** [#109](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/109) lands the adapter as `arctic_platform/integrations/trl/cortex.py` with 20 tests, stacked on #84. It's independent of this PR — needed for any Cortex run whichever encoding wins — so it can be reviewed on its own and doesn't add to the decision here.

The tests concentrate on the two failure modes that wouldn't be loud, since the zone rejects a badly-shaped envelope but happily trains on a mis-framed one: an off-by-one label frame, and a `loss_mask` one position wider than the labels. Both were confirmed discriminative by mutation rather than assumed — reverting either fix turns exactly its test red.

Its two real limits are enforced, not hidden: `temperature != 1.0` is refused in the constructor rather than mis-scaled, and entropy comes back as zeros, so entropy metrics on this path are meaningless and an entropy bonus is a silent no-op. GRPO as configured uses neither.

**In flight.** A recipe-scale run on the default image with no pinned tag, seed 7 plus the sign-flip control, replacing the debug-image caveat on the convergence table. vLLM is pinned to 0.27.1 so TRL imports with no stub — worth noting `trl` declares `vllm<=0.27.1` itself, so the 0.28 in my env was out of contract to begin with. I'll post the result either way.

The open question on auto-fallback vs. opt-in is unchanged and still yours.

sfc-gh-kganesan:
### Default-image recipe run landed — the debug-image caveat is retired

100 steps of GSM8K GRPO over Cortex on the **current default QA6 image, no pinned tag**, client-side loss with `client_loss_encoding="grpo"`, seed 7, plus the sign-flip control.

reward curves

| run | reward, first 20 steps | reward, last 20 | delta |
|---|---|---|---|
| client-side loss | 0.4007 | 0.6340 | **+0.2333** |
| advantages negated | 0.2166 | 0.0000 | **−0.2166** |

The control matters more than the headline number. It shares seed 7 and every hyperparameter with the normal run — I diffed the two arg dicts and they differ in exactly two keys, the output path and `negate_advantages`. So the only thing separating a run that climbs to 0.63 from one that collapses to zero by step 40 is the sign of the weights the client sends. That is the claim this PR makes, measured rather than argued.

Two secondary things the run settles:

**The log-prob frame is right.** Importance ratio stayed inside `[0.99937, 1.00244]` across both runs. This is the one check I said no client-side test could make — whether `compute_logprobs` and `compute_entropy_and_logprobs` agree on the frame — and on policy an off-by-one blows the ratio up rather than leaving it at 1.000. It didn't.

**No pinned image, no stub.** vLLM is at 0.27.1, which is what `trl` actually declares (`vllm<=0.27.1`); the 0.28 I'd been running was out of contract, and the `NCCLTrainerSendWeightsArgs` stub I briefly used is gone. The run went through [#109](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/109)'s adapter as a normal import, so this result is now reproducible from #84 + #95 + #100 + #109 with no scratch files.

`entropy: 0.0` throughout, as expected — #109 documents why, and GRPO as configured here doesn't read it.

Raw logs, metrics JSON and the plot are on `karthik/pr100-assets` under `assets/default-image/`.

Still yours: auto-fallback vs. landing as opt-in. Nothing about this run depends on that choice.

sfc-gh-kganesan:
### Handoff doc updated so you can reproduce this yourself

[`PR100-FOR-84.md`](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/karthik/pr100-assets/PR100-FOR-84.md) on `karthik/pr100-assets`. What's new, and two corrections you'd otherwise hit.

**Corrected: the merge instructions.** All three PRs — #100, #95, #109 — combine cleanly from current heads in either order, giving `98 passed`. But I first tried merging #109 into my older combined worktree and got a wall of add/add conflicts in files #109 never touches (`test_zorro_batch.py`, `integrations/trl/client.py`). Cause: both #100 and #95 were rebased onto newer `main` after that worktree was built, so the conflicts were pre-rebase artifacts, not real. The doc now says to build fresh and names the symptom, because it looks alarming and means nothing.

**Disclosed: my run did not use the current heads.** It used #100 at `1e7132c` (head is `e68cef1`), #95 at `bd0a61a` (head is `4dc4573`), and the adapter from a scratch file. I checked whether that invalidates the result rather than assuming:

- `git diff 1e7132c e68cef1 -- arctic_platform/integrations/trl/` is **empty**. The encoding is byte-identical to what ran.
- #95's transport changed by 35 lines, none on the forward path — `list_jobs`/`cancel_job`, richer terminal-state errors, a PAT accessor rename, empty-body handling for `:cancel`. Grepping that diff for `_WIRE_OPERATION|forward|logprob|octet|chunk` returns nothing.
- Scratch adapter vs #109 under `ast.unparse` with docstrings stripped: one extracted constant, two local renames, two error strings. No behavioural delta.

So it transfers, but a rerun on the merged head would be cleaner and is now one command.

**Added: everything needed to rerun.** Exact hyperparameters (Qwen3-1.7B, lr 3e-6, 8 generations, batch 32, 256-token completions, temperature 1.0), the vLLM 0.27.1 prerequisite, what a healthy run looks like by step 20 so you don't wait 2.5 hours to find out, and the expected trajectory with p-values (0.0052 within-run, <0.0001 against the control).

**Added: three metrics that will mislead you.** Worth reading before concluding something is broken:

1. `grad_norm` is `0.0` at every step **in both arms** — a logging artifact, since the optimizer step happens on Cortex and the local TRL trainer has no gradient to report. `learning_rate` logs fine. Don't use `grad_norm` as a health signal here.
2. The server's reported loss is not a loss. At ratio 1 the per-token GRPO term is `-advantages`, so it reports `sum(w)` rather than `sum(w·logprobs)`. Gradient right, value meaningless; the client reports its own loss to TRL and discards the server's.
3. `entropy` is `0.0` throughout, by construction.

**Fixed: a repro trap.** `harness/` had its own copy of the adapter, which would shadow #109 under the harness `PYTHONPATH` — a reproduction that silently tests a copy is worth nothing. Deleted; the recipe now imports `arctic_platform.integrations.trl.cortex`.

**Sharpened: what this does not establish.** The `5.2e-05` agreement is one batch. There's no 100-step server-side-loss run on seed 7 to overlay, so equivalence across a trajectory isn't shown — the honest framing is plumbing plus directional correctness with equivalence spot-checked at one step. Also n=1 seed on the default image, and neither #100 nor #109 gets CI while based on your branch, so every test count is from a local clean worktree.

Happy to run the server-side overlay if you want it before deciding.

sfc-gh-kganesan:
Reused the `grpo` encoding from this PR in #110, to carry Tinker's `cross_entropy` (which is what the SDK's `forward_backward_custom` builds its surrogate on). It transferred cleanly to a different integration, which is a decent sign the idea generalises.

One thing that might be worth folding back here: the ratio can be pinned to **exactly** 1.0 by *omitting* `old_log_probs_shifted` rather than setting it. Both `grpo_loss` in ArcticTraining-dss and AP's own do:

```python
old_log_probs_ctx = context.get("old_log_probs_shifted")
"old_log_probs": old_log_probs_ctx.to(...) if old_log_probs_ctx is not None else logprobs.detach(),
```

So the default is this same forward's log-probs, detached — not a recomputation. That removes the caveat in `_surrogate_payload`'s docstring about bf16 recompute leaving the ratio at 1 only to within forward nondeterminism (~1e-5 relative). It also saves sending a `[batch, seq]` tensor, which narrows the gap you note against the direct `weighted_logprob_sum` encoding.

Caveat on whether it applies to you: it holds only when the log-probs you'd send as π_old are the ones the *same* `fwd_bwd` recomputes. That is true in the Tinker path because the SDK's two passes are back-to-back with no `optim_step` between. If TRL ever sends π_old from an older policy snapshot, omitting it would silently turn an off-policy correction into an on-policy one — so this is only safe where π_old ≡ π_new is intended, which is exactly the surrogate's case but not the general one.

Agreed on treating the encoding as a stopgap. Registering `weighted_logprob_sum` in ArcticTraining-dss would retire it in both places at once; #110 already names that loss and only falls back because Cortex's registry lacks it.

PR #108, opened 2026-09-08
feat(openai_compat): OpenAI-compatible /v1 endpoint over a sampling job

sfc-gh-kganesan:
## What

An OpenAI-compatible `/v1` endpoint over an Arctic sampling job, so anyone evaluating a model we trained can keep the stack they already have — the `openai` SDK, LiteLLM, an eval harness — and change only `base_url`.

```bash
python -m arctic_platform.openai_compat --config client.json --port 8000
```

Non-streaming. `stream=true` returns a 400 rather than faking SSE deltas from an already-finished generation.

Four files: `translation.py` (pure — request models, parameter policy, OpenAI ⇄ sampling-job functions, no framework), `server.py` (routes, app, CLI), plus `__init__`/`__main__`. `client` is anything with `generate(prompts, sampling_params)`, sync or async.

## The rule it's built around

**A parameter that changes what the model should produce is either implemented or rejected, never silently ignored.** Accepting one and dropping it returns a plausible-looking 200 that is simply wrong, which is the most expensive failure mode to diagnose.

Applying that caught four bugs that would each have produced confidently wrong output with no error:

- **Content-part arrays** (`[{"type":"text",...}]`, which many clients emit for plain text) were `json.dumps`'d into the prompt, so the model saw a JSON blob and answered a different question.
- **`content: null`** on a tool-only assistant turn serialized to the literal string `null` in the transcript.
- **Omitted `max_tokens`** fell through to vLLM's `SamplingParams` default of **16 tokens**, truncating every reply from a client that doesn't set it.
- **`tools`** were dropped before the chat template, so the model was never told they existed and could never call one — an agent would look broken for reasons no log explains.

So `response_format`, `logit_bias`, and a forced `tool_choice` now 400 with the parameter named, while inert fields (`user`, `store`, `metadata`) are accepted and dropped. Full table in [the README](arctic_platform/openai_compat/README.md).

Also: errors use OpenAI's `{"error": {...}}` envelope (FastAPI's `{"detail": ...}` reaches the SDK as `Error code: 400` with the message gone); capacity errors relay as `429` + `Retry-After` instead of a `500` that ends the caller's run; a blocking client is driven through `asyncio.to_thread` so concurrent callers overlap.

Two operational choices: binding off-loopback without an API key is refused, since what makes the gateway reachable from a container also makes it reachable from anything else routing to the host. And the gateway never shuts the sampling job down — on Cortex `client.shutdown()` cancels the parent job and would take the endpoint's GPUs with it.

## Validated with a real eval harness on the full GSM8K test set

The strongest evidence here is an accuracy number, because a mangled prompt or a corrupted response collapses it. So: **EleutherAI's `lm-evaluation-harness` 0.4.13**, installed in its own venv, pointed at the endpoint with its stock `local-chat-completions` backend. It never touched a line of this code — it just spoke OpenAI to a real Cortex Qwen3-8B sampling job.

**Full GSM8K test set, all 1319 problems, 5-shot, 32 concurrent, 5m25s, zero failed requests:**

```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8863|±  |0.0087|
|     |       |strict-match    |     5|exact_match|↑  |0.8613|±  |0.0095|
```

88.6% is where Qwen3-8B is supposed to land, which is the point: the endpoint is not corrupting prompts, responses, stop sequences, or seeds anywhere in 1319 round trips.

The first full run actually scored **78.9%**, and chasing that produced a result worth keeping. It was a token budget, not a bug — Qwen3 thinks before answering, and `max_gen_toks=1024` truncates mid-`<think>`:

| `max_gen_toks` | flexible-extract | strict-match |
| --- | --- | --- |
| 1024 | 78.0% | 75.5% |
| 3072 | **87.5%** | **84.5%** |

*(200 problems, same endpoint, same job — only the token budget changed.)*

That is a ~10-point accuracy swing from one sampling parameter, and it is the empirical case for the `max_tokens` fix above. A client that sets the budget too low quietly loses ten points; a client that **omits** `max_tokens` entirely used to get vLLM's 16-token default and would have scored near zero, with a 200 on every request and nothing in any log to explain it.

## Live smoke test of the rest of the surface

Before the eval, a real Qwen3-0.6B sampling job on QA6, served through the CLI path, driven by the real `openai` SDK. **12/12**, job torn down cleanly afterwards.

```
job running after 239s: e0c3ab9b-...:sampling:0

  PASS  real generation          content="<think>\nOkay, the user is asking for the capital of France..."
  PASS  omitted max_tokens != 16 completion_tokens=485 finish=stop
  PASS  n=3                      3 choices
  PASS  stop + seed              stop honored, finish=stop
  PASS  logprobs                 8 tokens, first='<think>' lp=-0.000
  PASS  token_ids extension      prompt_token_ids[:5]=[151644, 872, 198, 13048, 151645]
  PASS  content parts
  PASS  tool calling             get_weather({"city": "Paris"}) finish=tool_calls
  PASS  legacy /v1/completions   text='Paris. The capital of France is also...'
  PASS  /v1/models               models=['qwen3-smoke']
  PASS  rejected param -> 400
  PASS  8 concurrent requests    8 concurrent requests in 0.8s
```

**A real model actually emitted a tool call** that came back in OpenAI shape with `finish_reason="tool_calls"` — the full round trip of tools rendered into the chat template, the model choosing to use one, and the parser recovering it. No stub can prove that.

## Offline suite

**67 tests, no GPU, ~5s.** `test_translation.py` covers the pure functions. `test_openai_compat.py` runs over real HTTP at two levels:

- Against a **stub that records its calls**, asserting on the *request the backend received* — a dropped sampling param, missing tools, or `max_tokens=16` are invisible in the response and obvious in the call. The tokenizer is the **real Qwen3 one**, because only a real `chat_template` proves tools reach the prompt.
- Against the **real `openai` SDK on a real socket**. The SDK parses every response into OpenAI's own generated Pydantic models, so a wrong-shaped response raises inside the client — making this a conformance check against OpenAI's schema rather than against my idea of it. Errors are asserted as `openai.BadRequestError`, `RateLimitError`, `NotFoundError`, `AuthenticationError`.

Also confirmed out-of-process with **LiteLLM 1.101.0rc1** from a separate venv (8/8, including a tool-call round trip and the `hosted_vllm` provider path, whose stricter `model_info` rules some harnesses take).

So the endpoint has now been driven by **four independent clients** — the `openai` SDK, LiteLLM, lm-evaluation-harness, and curl — none of which know anything about Arctic Platform.

## Three gotchas, documented rather than discovered later

Each found by running a real client's code, not by guessing:

1. **Model naming.** LiteLLM's `hosted_vllm` provider requires exactly one `/` and `[A-Za-z0-9._-]{1,64}`, so `hosted_vllm/Qwen/Qwen3-8B` is rejected before a request is sent. Use `--served-model-name qwen3-8b`.
2. **Reasoning models.** Qwen3 has thinking on by default; the `<think>` block lands in `content` and counts against `max_tokens` — the GSM8K table above is what that costs. vLLM's server can split it into `reasoning_content`; this endpoint doesn't. Budget for it, or pass `chat_template_kwargs={"enable_thinking": false}`.
3. **Token-id prompts.** `/v1/completions` accepts them and Cortex handles them, but the on-prem server validates `prompts` as `list[str]` and rejects them. Chat is unaffected.

## Notes for review

- Nothing outside the new package changes except one `pyproject.toml` extra (`arctic_platform[openai]` — serving deps only; it drives a sampling job through the client, so it pulls no engine).
- The second commit cuts the first one down: I'd shipped an in-process threaded server class nobody asked for, and a 138-line `backend.py` whose job was to choose between `await` and `to_thread`. Product Python is 819 lines.
- Streaming was left out deliberately rather than stubbed. `/v1/completions` is ~85 lines of the total and could go too if we only care about chat.


sfc-gh-thonguyen:
Reviewed head `e6c9504`. I have several concurrency and compatibility concerns before relying on this adapter for high-concurrency evals.

### Concrete findings

1. **The CLI concurrency flag does not configure the actual backend execution capacity.** The CLI constructs a blocking `ArcticClient`, and [`_generate()`](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/server.py#L100) dispatches it through `asyncio.to_thread()`. On Python 3.12, the default executor has at most 32 workers. Each worker remains occupied throughout generation, including Cortex polling. In a local stub check with 96 concurrent requests and `max_concurrency=128`, backend concurrency still peaked at 32; an async stub reached 96. This is a configurable default, not a hard Python limit, but raising `--max-concurrency` alone does not remove it. Could the CLI use the existing `AsyncArcticClient`, or explicitly configure the executor alongside the semaphore, with a test above 32 concurrent calls?

2. **Prompt arrays are processed serially.** The [`/v1/completions` loop](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/server.py#L209) awaits each prompt before submitting the next. Eight prompts against a 100ms async stub took 801ms, with peak backend concurrency of one. Could compatible prompts be passed through as a backend batch, or at least dispatched with bounded concurrency while preserving response ordering? The current behavior defeats batching for eval clients sending prompt arrays.

3. **Async backend errors lose their HTTP status and retry guidance.** [`_as_openai_error()`](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/server.py#L84) reads `exc.response.status_code`, but the supported async client uses `aiohttp.ClientResponseError`, whose status and headers are directly on the exception. I confirmed that an aiohttp 429 becomes a 502 with no `Retry-After`. Please handle both exception shapes and test the actual aiohttp shape, especially if switching the CLI to async.

4. **`top_logprobs` is accepted but the alternatives are discarded.** [`_logprobs()`](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/translation.py#L350) always emits `top_logprobs: []`, even when the backend supplies candidate probabilities. The legacy response similarly emits `top_logprobs: None`. Could we preserve the requested alternatives, or reject unsupported requests instead of returning a successful response missing the requested data? An assertion on the returned alternatives would catch this; the current tests only check sampled-token probabilities.

### Additional scaling concerns

- **CPU work remains on the event loop.** [Chat templating and tokenization](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/server.py#L172), plus response conversion and per-token decoding, execute synchronously in the async handlers. The CLI starts one Uvicorn worker. Long prompts or logprob-heavy responses can therefore delay unrelated requests even after backend dispatch becomes async.
- **The semaphore bounds active calls, not the waiting queue.** [Admission](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/e6c9504a27a408b361a64fb2f9881a849066bebc/arctic_platform/openai_compat/server.py#L60) has no explicit pending-request bound or queue deadline. For sustained overload, could we add bounded admission and a clear overload response rather than allowing queue latency and retained request memory to keep growing?

The extra HTTP hop is not inherently a bottleneck, and the backend can still continuously batch admitted requests. However, compatibility validation and throughput validation are different: the reported 32-concurrent GSM8K run does not exercise the executor ceiling or establish behavior under higher load. A direct-client versus adapter comparison at increasing concurrency, with throughput, p95 latency, queue time, event-loop lag, and GPU utilization, would help distinguish adapter limits from backend capacity.

**Validation:** All 67 offline tests passed locally. The concurrency, serial-batch, async-error, and logprob observations above were checked with local stubs or translation calls; I did not run a live GPU throughput comparison.

sfc-gh-kganesan:
Thanks — this was a genuinely useful review. All four findings were real, and they're fixed in `6b1c5e1`. I reproduced each one as a test first and confirmed the tests fail on `e6c9504`, so the gaps you identified in the existing coverage are now closed rather than just patched.

### The four findings

**1. The concurrency flag didn't configure execution capacity.** Correct, and your diagnosis of the mechanism was exact. `_State` now owns a `ThreadPoolExecutor` sized to `max_concurrency` instead of riding the default executor, and the CLI builds an `AsyncArcticClient`, so an in-flight request costs a coroutine rather than a pooled thread. Test asserts peak backend concurrency above 32 with the flag at 48.

**2. Prompt arrays were serial.** Fixed with `asyncio.gather`, which preserves response order while the semaphore still bounds what's in flight. I used `return_exceptions=True` and re-raise the first failure, so one bad prompt doesn't leave the survivors running as orphaned backend work. I didn't pass the array through as a single backend batch — per-prompt `max_tokens` resolution depends on each prompt's own length, so they aren't always one homogeneous request — but they now dispatch concurrently, which recovers the batching you were after.

**3. Async errors lost their status.** Correct, and worth stressing that this was a prerequisite for fixing 1 rather than an independent nit: going async is exactly what makes the aiohttp path live. `_as_openai_error` now reads status and headers from either shape, and the test raises the aiohttp shape specifically — status and headers on the exception — and asserts both the 429 and the `Retry-After` value survive.

**4. `top_logprobs` discarded the alternatives.** You're right, and this one was the most embarrassing: the requested count was already being forwarded to the sampler as `params["logprobs"]`, so the candidates were arriving on the wire and getting dropped at the final translation step. It's also precisely the failure mode every other unsupported field in this adapter is rejected to avoid — a plausible 200 that silently lacks what was asked for. Now ranked, truncated to the requested `k`, and the legacy endpoint emits the token→logprob maps its schema calls for instead of `None`.

### Throughput comparison

Two independent runs, Qwen3-8B on one GPU, same sampling job and same client type on both sides, so the delta is the adapter: HTTP hop, chat templating, response translation. Loop lag is sampled inside the gateway's own event loop. Both numbers shown are run 1 / run 2.

| concurrency | direct req/s | adapter req/s | direct p95 | adapter p95 | loop lag p95 |
|---|---|---|---|---|---|
| 1 | 0.72 / 0.73 | 0.74 / 0.76 | 2014 / 1915 ms | 1844 / 1819 ms | 0.4 / 0.3 ms |
| 8 | 4.81 / 7.34 | 5.20 / 5.47 | 1854 / 1118 ms | 1623 / 1727 ms | 0.6 / 0.8 ms |
| 32 | 21.8 / 21.9 | 26.1 / 18.2 | 1716 / 1506 ms | 1471 / 1521 ms | 0.9 / 0.8 ms |
| 64 | 43.0 / 41.7 | 35.8 / 41.8 | 1864 / 1432 ms | 2034 / 2089 ms | 1.1 / 5.2 ms |
| 128 | 62.1 / 67.2 | 58.2 / 49.7 | 2110 / 2152 ms | 2925 / 3093 ms | 4.6 / 12.0 ms |

Reading these honestly: run-to-run variance swamps the middle of the table. Concurrency 32 came out 1.20x one run and 0.83x the next, so I wouldn't defend any single-digit-percent claim there — the measurement windows are only 2 to 5 seconds at high concurrency on a shared QA6 backend. What reproduced in both runs is the top row: at 128 concurrent the adapter costs throughput and adds roughly 800–950 ms at p95. The part that matters for finding 1 is that both sides now scale past 60 req/s, which the old build could not have done at any flag setting.

I can't give you GPU utilization — it isn't observable from the client side of a Cortex job.

### Your two scaling points

Both are fair and I'd rather answer than quietly half-implement them.

**CPU on the event loop** is real and now measured: lag grows with concurrency, to 12 ms p95 at 128 concurrent. Against ~1700 ms request latency that's under 1%, so it isn't the binding constraint yet, but the trend is exactly what you predicted and it will matter with longer prompts or heavier logprob traffic. Moving templating off the loop is the obvious next step if that shows up.

**Bounded admission** I've deliberately left out of this change. An explicit pending bound and overload response is a design addition with its own failure modes — what the queue depth should be, whether to shed or block, what a harness does with the resulting error — and folding it into a review-response commit is how it ends up untested. Happy to take it as a follow-up if you think it should land before this is used for sustained high-concurrency work.

Suite is 73 passing, lint and license clean.

PR #109, opened 2026-09-10
Cortex envelope adapter for the TRL integration

sfc-gh-kganesan:
Stacked on #84. Small: one module, one test file.

## Why

The TRL integration speaks the on-prem server's dialect. It sends `{batch, meta, processing}` and asks for the post-processors `apply_temperature` and `compute_entropy_and_logprobs`. A Cortex zone wants `{kwargs, processing}` and registers only `compute_logprobs` — the other two names appear nowhere in ArcticTraining-dss or dss-platform.

So on Cortex the recipe dies before it reaches the loss at all:

```
dispatch requires 2D input_ids to plan token-budget microbatches
```

from the zone's `dp_shard_batch`, which is being handed the outer envelope instead of a batch.

`CortexTRLAdapter` sits between `ArcticTrainingClient` / `ArcticRolloutWorker` and the real client and translates. Only `fwd_no_grad` and `fwd_bwd` carry a batch envelope; everything else passes through untouched.

## Why now, and why it isn't hiding in a scratch file anymore

This translation is what every end-to-end result on the Cortex path has quietly depended on, including the three-seed GSM8K convergence run. It lived in my scratch directory, which meant that run could not be reproduced from the PRs — anyone starting from #84 + #95 would hit the `dispatch requires 2D input_ids` wall. That's the gap this closes. It is load-bearing infrastructure, so it belongs in the tree where it can be reviewed and tested.

Worth being clear that this is a translation layer, not the end state. Two things retire it: registering `apply_temperature` and `compute_entropy_and_logprobs` in ArcticTraining-dss, or teaching the integration to emit both dialects directly. I'd rather have the second, but either beats a shim, and neither is a reason to keep the Cortex path unrunnable in the meantime.

## The two gaps it can't close

Both are enforced or documented rather than papered over:

| gap | what the adapter does |
|---|---|
| `apply_temperature` doesn't exist on the zone | refuses any `temperature != 1.0` in the constructor, where it's a no-op. Mis-scaled logits would otherwise be invisible. |
| `compute_logprobs` returns no entropy | returns zeros to keep the caller's shape contract. **Any entropy metric on this path is meaningless and an entropy bonus is a silent no-op.** GRPO as configured uses neither, which is the only reason it's survivable. |

If a zone ever does return entropy, it's passed straight through — there's a test pinning that, so the zeros stop the moment they're no longer necessary.

One thing this file cannot assert: whether `compute_logprobs` and `compute_entropy_and_logprobs` agree on the *frame* of the log-probs they return. That's a property of two server implementations, not of the translation. The observable check is the importance ratio, which must sit at ~1 on policy; an off-by-one frame blows it up. It did sit at ~1 across the seed runs.

## Tests

20 tests. Most of the surface fails loudly if mistranslated — the zone rejects a bad envelope — so the tests concentrate on the two failures that would be *quiet*, because both produce a running job that optimises the wrong thing:

- **an off-by-one label frame.** `compute_logprobs` scores position `t` against `labels[t]`, so labels are next-token and a row's final real token is not a target.
- **a `loss_mask` one position wider than the labels.** Callers derive it from `attention_mask`, which is one position too generous once labels exist; grpo's preflight requires `loss_mask == 0` wherever the label is `-100`.

Both were checked by mutation rather than assumed: reverting either fix turns exactly its test red, and dropping the unknown-post-processor guard turns its test red too.

## What I did not verify here

These are unit tests against a recording client, so they pin the translation and nothing about the zone. The end-to-end evidence is separate: `fwd_no_grad` over `CortexTransport` on the current default QA6 image, gradient agreement to 5.2e-05 against the direct grpo path, and the seed runs. A default-image recipe-scale run is in flight and I'll post it when it lands.

### DIFF
--- a/arctic_platform/integrations/trl/cortex.py
+++ b/arctic_platform/integrations/trl/cortex.py
@@ -1,0 +1,154 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Make this integration's client calls legible to a Cortex zone.
+
+The TRL integration was written against the on-prem server and speaks its
+dialect: batches go out as ``{"batch", "meta", "processing"}`` and every call
+asks for the post-processors ``apply_temperature`` and
+``compute_entropy_and_logprobs``. A Cortex zone understands neither. It wants
+``{"kwargs", "processing"}``, and the only post-processor it registers is
+``compute_logprobs`` -- the other two names do not appear anywhere in
+ArcticTraining-dss or dss-platform.
+
+Without translation the recipe fails before it ever reaches the loss: the zone's
+``dp_shard_batch`` rejects the outer envelope with ``dispatch requires 2D
+input_ids to plan token-budget microbatches``.
+
+:class:`CortexTRLAdapter` sits between :class:`ArcticTrainingClient` /
+:class:`ArcticRolloutWorker` and the real client, and translates.
+
+This is a translation layer, not the end state. The two candidates for a real
+fix are registering those post-processors in ArcticTraining-dss, or teaching
+this integration to emit both dialects directly. Either retires this file. It
+exists because the alternative is that the Cortex path cannot run at all.
+
+Two gaps it cannot close, both raised or documented rather than hidden:
+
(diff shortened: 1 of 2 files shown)


sfc-gh-kganesan:
### Exercised at recipe scale on the default image

The unit tests here pin the translation and say nothing about the zone, so: this adapter just carried 100 steps of GSM8K GRPO over Cortex on the current default QA6 image, twice.

reward curves

Reward went 0.401 → 0.634 on seed 7, and the same seed with advantages negated went 0.217 → 0.000. Both ran end to end with the adapter imported normally rather than injected from a scratch directory, which was the point of the PR.

The result worth calling out for *this* diff is the importance ratio: `[0.99937, 1.00244]` across both runs. The module docstring says the one thing the adapter cannot self-assert is whether `compute_logprobs` and `compute_entropy_and_logprobs` agree on the frame of the log-probs, and that the observable proxy is a ratio pinned at ~1 on policy. It is. So the `next_token_labels` framing agrees with what the on-prem post-processor would have produced, at least in the on-policy regime this recipe stays in.

`entropy: 0.0` for all 200 logged steps, exactly as documented. GRPO as configured reads neither entropy nor a bonus, so it cost nothing here — but it is still the reason I would not run an entropy-regularised recipe over this path.

Logs and metrics: `assets/default-image/` on `karthik/pr100-assets`.

PR #110, opened 2026-09-11
tinker: run tinker-cookbook recipes on Cortex Training

sfc-gh-kganesan:
Serves Tinker's HTTP API over Cortex Training, so an **unmodified**
`tinker-cookbook` recipe trains on Cortex. The recipe does not change at all —
you point it at a different `base_url`.

Verified against `tinker==0.25.0` / `tinker-cookbook==0.5.5`.

```
tinker-cookbook recipe  ──HTTP──►  serve.py (laptop / CPU VM)  ──►  Cortex job (GPUs)
      unchanged                    speaks Tinker, talks Cortex
```

`serve.py` is a translator, not a trainer: no local GPU. Provisioning lives
there because Tinker's protocol has no verb for "give me two GPUs with ZeRO-2".

> **Base is #95, not `main`** — it needs the no-grad forward from there to back
> Tinker's `forward` verb. Retarget once #95 lands.

---

# Quickstart

**1. Install** (nothing GPU-related; the server needs `tinker` only for its
generated protobuf schema)

```bash
pip install "arctic_platform[cortex]"
pip install "tinker==0.25.0" "tinker-cookbook==0.5.5"
```

**2. Point at a Cortex account**

```bash
export ARCTIC_CORTEX_HOST=<account>.<region>.snowflakecomputing.com
export ARCTIC_CORTEX_DATABASE=<db>
export ARCTIC_CORTEX_SCHEMA=<schema>
export ARCTIC_CORTEX_PAT=<your PAT>
```

`CortexConfig` is a `pydantic-settings` model, so these populate it directly;
`--config conn.json` reads the connection from a file instead.

**3. Start the server**

```bash
python -m arctic_platform.integrations.tinker.serve \
    --model Qwen/Qwen3-0.6B \
    --training-gpus 1 --sampling-gpus 1 \
    --max-prompt-length 1024 --max-response-length 512 \
    --zero-stage 2 --port 8112
```

This takes 3–4 minutes and **looks hung** — Cortex is placing the GPUs, and it
takes considerably longer on a busy cluster. Ready when this answers:

```bash
curl -s http://127.0.0.1:8112/api/v1/get_server_capabilities
# {"supported_models":[{"model_name":"Qwen/Qwen3-0.6B"}]}
```

`--job-id <id>` attaches to an already-running job and leaves it running on
shutdown. Without it, stopping the server releases the GPUs.

**4. Run a recipe**

```bash
TINKER_API_KEY=tml-dummy python -m tinker_cookbook.recipes.math_rl.train \
    base_url=http://127.0.0.1:8112 \
    model_name=Qwen/Qwen3-0.6B \
    renderer_name=qwen3_disable_thinking \
    lora_rank=0 \
    env=gsm8k group_size=8 groups_per_batch=8 \
    max_tokens=384 temperature=1.0 learning_rate=2e-6
```

## Arguments that are not free choices

Three of the above are constraints, and all three fail **before** any network
request, which makes them confusing the first time:

| argument | why |
|---|---|
| `lora_rank=0` | Cortex rejects `rank > 0` with a 400. The cookbook defaults to 32. |
| `learning_rate=2e-6` | The cookbook's default `1e-5` is calibrated for LoRA. Forced into full fine-tuning it destroys a 0.6B — see the collapse curve below. |
| `TINKER_API_KEY=tml-` prefix | The SDK validates the prefix client-side. The value is unused. |
| `renderer_name=...` | Required under 4B: `get_recommended_renderer_name` consults a hardcoded table starting at 4B and raises `KeyError` before any request. `math_rl` takes the name directly. |

## The metric to watch

`kl_sample_train_v1` should sit in **0.0005–0.005**. It is the divergence
between what the sampler produced and what the trainer thinks it produced, so
it is the alarm for frame and alignment bugs. Above ~0.05 is plumbing, not
hyperparameters.

**A healthy KL attests to the transport, not to the run.** In the collapsed
`1e-5` run below, KL *fell* to 0.0005 while the policy destroyed itself. Read
KL for plumbing; read `correct` and `ac_tokens_per_turn` for learning.

---

# Limitations

## Hard blocks

| limitation | detail |
|---|---|
| **No `cispo` / `dro`** | Both return 400. `cross_entropy` *is* supported — see "Custom losses" below. |
| **Full fine-tuning only** | LoRA `rank > 0` returns 400. This is what makes the cookbook's LoRA-calibrated default LR unsafe. |
| **`temperature=1.0` only** | Cortex registers only `identity` and `compute_logprobs` as post-processors, so there is no `apply_temperature` and the trainer scores at 1.0 whatever the sampler did. `sample` now **returns 400** for any other temperature instead of letting the two diverge silently. |
| **No multimodal** | `ModelInput` accepts `EncodedTextChunk` only; image, DMEL, and asset-pointer chunks are refused. |
| **No checkpoint resume** | `load_weights` is unserved, so `load_checkpoint_path` 404s. `save_weights` works. |
| **Clipping is inert by construction** | Cortex re-derives π_old from the live forward rather than a rollout snapshot, and the cookbook takes one optimizer step per batch, so π_old ≡ π_new and `approx_kl` reads exactly 0.0. Confirmed live: `importance_weight` 1.0, `clip_ratio` 0.0, `avg_loss` == −mean(advantage). A recipe relying on several updates per batch cannot be reproduced. |
| **Single training run, no auth** | As in #43. |
| **`dp_size` pinned to 1** | Multi-DP would need per-shard chunking in the adapter. Untested. |

## Custom losses

`forward_backward_custom` works. It is worth explaining how, because the
mechanism is not obvious and it constrains what the server's metrics mean.

The verb is entirely client-side: the SDK runs a `forward`, differentiates your
loss locally under autograd, and sends `w = dC/dlogprobs` back as cross-entropy
`weights`, relying on backend CE being `sum(-logprobs * w)` so the chain rule
reconstitutes your loss. The surrogate is hardcoded to `cross_entropy` and
cannot be configured — which is why gating that name off previously killed the
verb on its first call.

Cortex registers no such loss, so the binder encodes it through stock `grpo`:

- grpo's gradient wrt log-probs is `-advantages * ratio`, so `advantages = -w`
  leaves exactly `w`.
- **Omitting** `old_log_probs_shifted` makes grpo default π_old to
  `logprobs.detach()` — the *same* forward's output, not a second forward's — so
  the ratio is exactly 1.0 and clipping provably cannot engage. This is where
  this differs from #100, which sends π_old from a separate client forward and
  so lands at ratio 1 only to within forward nondeterminism (~1e-5 relative).
  Omitting it is both exact and free.
- `batch_num_tokens=1` cancels grpo's token-mean divisor, since Tinker's CE is
  an unnormalized sum and the client has already scaled `w`.

Two consequences a reviewer should hold onto: **server-side loss metrics are
meaningless on this path** (the reported loss is `sum(-advantages)`, and the SDK
discards it in favour of the client's own), and **so is any entropy metric**,
because grpo computes `entropy = -logprobs.detach()` rather than a real entropy.

Nothing downstream reads the loss value, so a sign or frame error here does not
raise — it trains the wrong way. The verification is sized to that:
`TestGradientEquivalence` asserts `dL/dlogprobs == -weights` through the whole
pipeline and includes a negation-dropped control so the assertion cannot pass
vacuously; on a live job, client-side CE fell 2.2032 → 0.0000, with three rows
of differing length each descending independently, which is the only way a
per-row frame shift surfaces.

This also fixed a real bug in `forward`: the final target token was appended
only for `forward_backward`, so pass 1's last log-prob scored a pad. Harmless
for RL, silently wrong for a client differentiating those log-probs.

## Operational gotchas

- **A job's weights persist and there is no reset verb.** Every run against the
  same job inherits the previous run's weights, so a baseline number is
  meaningless without recycling the job. This cost me real time: a "collapse" I
  diagnosed as a learning-rate problem was partly inherited damage from a prior
  run.
- **Provisioning is 3–4 min at 1+1 GPUs** and much longer at larger shapes. Job
  states include `PLACING`, which is easy to miss when filtering for busy jobs —
  GPUs look free when they are not.
- **Releasing is not instant.** A launch within ~a minute of a cancel can still
  hit the per-account GPU cap.
- **The published Cortex API spec is stale** on `post_process_outputs`. The live
  shapes are pinned in `TestCortexResponseShapes`.
- **Cookbook model table starts at 4B**, so small models need a recipe that
  accepts `renderer_name` explicitly.

## Not verified

Anything above 1+1 GPUs (larger shapes sat in `PLACING` for 12–25 min on the
shared cluster) and multi-GPU data parallel.

Custom losses are verified by gradient equivalence and by memorising a fixed
batch, not by a full SFT recipe run — the cookbook's SFT recipes also want
`load_weights`, which is unserved.

`ppo` and `importance_sampling` are indistinguishable on this path. Cortex
re-derives π_old from the live forward, so the ratio is always 1 and the clip
thresholds never bind; `meta["actor_config"]` is also dropped by
`to_cortex_fwd_bwd_payload`, so Tinker's `clip_low_threshold` /
`clip_high_threshold` never reach the server at all. Inert today for the
one-step-per-batch loop the cookbook runs, but it would matter to a recipe doing
several updates per batch.

---

# How to review

`+3,718` overstates it. `router.py` is carried from #43 and moved, so git shows
all 1,142 lines as new when only ~250 changed:

| file | | size |
|---|---|---|
| `cortex.py` | **new** — read first, it is the whole integration | 324 |
| `proto_wire.py` | **new** | 238 |
| `serve.py` | **new** — provisioning + CLI, mostly config plumbing | 185 |
| `router.py` | carried from #43, **+215 −115** against that branch | 1,205 |
| `test_cortex_binder.py`, `test_proto_wire.py`, `test_custom_loss.py` | **new** | 700 |
| `conftest.py`, `test_adapters.py`, `test_wire_schema.py`, `test_tinker_router.py` | carried, ~±140 total (repointed imports, plus the custom-loss and temperature cases) | 1,157 |

Review surface is roughly **1,240 new lines plus ~250 changed in `router.py`**.
To see just the router delta:

```bash
git diff origin/karthik/tinker-http-layer:arctic_platform/rl/tinker_router.py \
         origin/karthik/tinker-cortex:arctic_platform/integrations/tinker/router.py
```

The seven commits are independently reviewable and in dependency order: bind →
proto wire → Cortex conventions → the gradient fix → docs → comment cleanup →
custom loss + temperature refusal.
`arctic_platform/integrations/tinker/README.md` is the design doc.

## Three things worth your attention

**1. The wire is protobuf, not JSON** (`proto_wire.py`). Current SDKs post
`forward_backward` as `application/x-protobuf` and *reject* a JSON reply for
`ForwardBackwardOutput` and `SampleResponse` ("...which this SDK version only
supports as proto"). A JSON-only server fails at the first training step no
matter how correct its numbers are. #43's schema tests could not catch this
because they compared the router against hand-copied schema types rather than
the SDK; `test_proto_wire.py` uses the SDK's own `request_conv` /
`response_conv` as the oracle in both directions.

**2. Two conventions fail silently rather than raising.** Each has a dedicated
test.

- *Row alignment* (`cortex.py`). The router lays rows out as
  `[pad… prompt][response pad…]`; Cortex's packer needs real tokens in the
  leading columns. Aligning alone is not enough — log-probs come back in the
  aligned frame while the router's slices index the original one, so the
  permutation is inverted on the way back. Skipping the inverse shifts each row
  by its own prompt padding, with no error.
  `test_skipping_the_inverse_would_shift_rows` fails if the inverse is removed.
- *The final target token* (`router.py`). `target_tokens[-1]` is not present in
  `model_input`, so without appending it the last scored position has nothing to
  predict — and the advantage sitting there still multiplies its gradient. This
  was a real bug and surfaced **only** as `kl_sample_train_v1` reading
  0.35–0.49; it is 0.001–0.005 after the fix.

**3. `_require_logprobs` raises instead of defaulting** (`cortex.py`). The
router's fallback is an empty `loss_fn_outputs`, which the cookbook hits as a
bare `KeyError` frames away. Failing loud is what located Cortex's
`post_process_outputs` shape.

---

# Evidence

A full 935-step GSM8K epoch (`done_frac` 1.0), Qwen3-0.6B, 1 training + 1
sampling GPU, via the Quickstart command verbatim:

| steps | 0–115 | 116–231 | 232–347 | 348–463 | 464–579 | 580–695 | 696–811 | 812–927 |
|---|---|---|---|---|---|---|---|---|
| `correct` | 0.631 | 0.699 | 0.725 | 0.738 | 0.708 | 0.726 | 0.718 | 0.740 |
| `ac_tokens_per_turn` | 221 | 190 | 174 | 188 | 190 | 196 | 215 | 245 |
| `kl_sample_train_v1` | 0.00110 | 0.00086 | 0.00084 | 0.00158 | 0.00160 | 0.00160 | 0.00259 | 0.00353 |

63% → 74% correct, with responses getting *shorter* while accuracy rises
through step ~347 — the direction matters, since a model gaming the reward gets
longer. Plateaus after ~230 steps: converged for this configuration, not
stopped while climbing.

The same recipe at the cookbook's default `learning_rate=1e-5`, also a full
epoch, for contrast:

| steps | 0–115 | 116–231 | 232–347 | 348–463 | … | 812–927 |
|---|---|---|---|---|---|---|
| `correct` | 0.424 | 0.098 | 0.013 | 0.000 | | 0.000 |
| `ac_tokens_per_turn` | 344 | 384 | 384 | 384 | | 384 |
| `kl_sample_train_v1` | 0.00092 | 0.00146 | 0.00075 | 0.00052 | | 0.00048 |

Responses grow until every rollout hits the token cap and is truncated before
stating an answer. Correctness anti-correlates with length at **−0.635** here
against −0.321 in the healthy run.

---

# How this was validated

Five layers, each catching what the one below could not. The server ran in one
env and the SDK/cookbook client in another, so nothing passed by sharing
in-process objects.

1. **Unit tests** — 103, no GPU and no Cortex account. Two are adversarial
   rather than confirmatory: `test_proto_wire.py` uses the SDK's own converters
   as oracle, and `test_skipping_the_inverse_would_shift_rows` fails if the
   un-align step is deleted. The binder's stub echoes `input_ids` back as
   log-probs so a frame shift shows up as wrong token values, not plausible
   floats.
2. **Real SDK against a mock server** — the router with stub handlers, driven by
   the genuine SDK through every verb, with no GPU involved. This is the layer
   that found the protobuf wire, plus the `tml-` prefix check and the `encoding`
   oneof name.
3. **Shape probes against a live Cortex job** — established the
   `forward` vs `forward-backward` log-prob shapes, that `generate` has no `n`
   and keys log-probs by token-id string, and that `compute_logprobs` uses the
   predicting-position frame.
4. **A sampler-vs-trainer correctness probe** — replicates the cookbook's datum
   convention, samples a real completion, feeds it back as a training datum and
   compares log-probs at shifts −1/0/+1. This is what found the final-target-token
   bug: `mean|diff|` 0.845 → 0.043 after the fix. **Layers 1–3 all passed while
   that bug was live**, because shapes and wire were correct and every verb
   returned plausible numbers.
5. **Full epochs through the unmodified cookbook** — several runs; the sequence
   of failures (wrong `max_tokens`, inherited weights, LoRA-calibrated LR) is
   what established the constraint list above.

## Test plan

- [x] `pytest tests/integrations/tinker` — 103 pass, no GPU or Cortex account
- [x] `pytest tests/integrations tests/client` — 197 pass
- [x] Two full 935-step GSM8K epochs against live Cortex jobs (above)
- [x] All 12 verbs round-trip against the real SDK over the proto wire
- [x] Proto codec round-trips against the SDK's own converters
- [ ] Reviewer sanity-check of the loss mapping (`ppo` / `importance_sampling`
      → Cortex `grpo`)

### DIFF
--- a/arctic_platform/integrations/tinker/README.md
+++ b/arctic_platform/integrations/tinker/README.md
@@ -1,0 +1,291 @@
+# Tinker integration
+
+Serve [Tinker](https://tinker-docs.thinkingmachines.ai/)'s HTTP API over Arctic
+backends, so an **unmodified** [`tinker-cookbook`](https://github.com/thinking-machines-lab/tinker-cookbook)
+recipe trains on Cortex Training. The recipe believes it is talking to Tinker;
+`TINKER_BASE_URL` is the only thing that changes.
+
+Verified against `tinker==0.25.0` / `tinker-cookbook==0.5.5`.
+
+## Layout
+
+| File | Role |
+|---|---|
+| [`router.py`](./router.py) | Tinker's protocol as a FastAPI router. Backend-agnostic: `init_tinker_state` injects five handlers and the router knows nothing else. Owns sessions, futures, the datum/batch conversion and the loss-name mapping. |
+| [`proto_wire.py`](./proto_wire.py) | The protobuf codec the modern SDK requires (§5). Imports `tinker.proto.tinker_public_pb2` lazily, so the router still works without the SDK installed. |
+| [`cortex.py`](./cortex.py) | Binds the five verbs to `AsyncArcticRLClient`. Owns the envelope dialect and the row-alignment round trip (§6). |
+| [`serve.py`](./serve.py) | Provisions the Cortex job from flags and serves the app. Releases the job on shutdown unless `--job-id` attached it to someone else's. |
+
+The router being a pure adapter is what keeps this small: the Tinker surface
+reaches Cortex without the on-prem server, DeepSpeed, or Ray being involved at
+all. Provisioning is deliberately outside Tinker's protocol — there is no Tinker
+verb for "give me two GPUs with ZeRO-2 and FA3", so `serve.py` takes that from
+flags and binds the Tinker surface onto the result.
+
+## 1. Install
+
+```bash
+pip install "arctic_platform[cortex]"
+pip install "tinker==0.25.0" "tinker-cookbook==0.5.5"
+```
+
+The server needs `tinker` only for its generated protobuf schema (§5). It does
+not need a GPU — Cortex owns those.
+
+## 2. Set Cortex env
+
+```bash
+export ARCTIC_CORTEX_HOST=<account>.<region>.snowflakecomputing.com
+export ARCTIC_CORTEX_DATABASE=<db>
+export ARCTIC_CORTEX_SCHEMA=<schema>
+export ARCTIC_CORTEX_PAT=<your PAT>
+```
+
+`CortexConfig` is a `pydantic-settings` model, so these populate it directly.
+Pass `--config conn.json` instead to read the connection from a file.
+
+## 3. Run the server
+
+```bash
+python -m arctic_platform.integrations.tinker.serve \
+    --model Qwen/Qwen3-0.6B \
+    --training-gpus 1 --sampling-gpus 1 \
+    --max-prompt-length 1024 --max-response-length 512 \
+    --zero-stage 2 --port 8112
+```
+
+Expect ~3–4 min before it answers: the Cortex job has to be placed. Startup is
+complete when `get_server_capabilities` returns:
+
+```bash
+curl -s http://127.0.0.1:8112/api/v1/get_server_capabilities
+# {"supported_models":[{"model_name":"Qwen/Qwen3-0.6B"}]}
+```
+
+`--job-id <id>` attaches to an already-running job instead of creating one,
+and then leaves it running on shutdown. Without it, stopping the server
+releases the GPUs.
+
+## 4. Point a recipe at it
+
+```bash
+TINKER_API_KEY=tml-dummy python -m tinker_cookbook.recipes.math_rl.train \
+    base_url=http://127.0.0.1:8112 \
+    model_name=Qwen/Qwen3-0.6B \
+    renderer_name=qwen3_disable_thinking \
+    lora_rank=0 \
+    env=gsm8k group_size=8 groups_per_batch=8 \
+    max_tokens=384 temperature=1.0 learning_rate=2e-6
+```
+
+Three of those arguments are not free choices, and each is a constraint from
+§7 rather than a tuning preference:
+
+* **`TINKER_API_KEY` must start with `tml-`.** The SDK validates the prefix
+  client-side before any request. The value is otherwise unused here.
+* **`renderer_name` must be explicit** for models below 4B.
+  `get_recommended_renderer_name` looks the model up in a hardcoded table that
+  starts at 4B, so a 0.6B raises `KeyError` before any request is sent.
+  `math_rl` accepts the name directly, which short-circuits the table.
+* **`lora_rank=0`** — Cortex rejects `rank > 0` with a 400, so this path is
+  full fine-tuning only. That has a consequence for the learning rate, below.
+
+## 5. The protobuf wire
+
+Modern SDKs post `forward_backward` as `Content-Type: application/x-protobuf`
+and **reject a JSON reply** for `ForwardBackwardOutput` and `SampleResponse`:
+
+```
+Server returned a JSON payload for {model_cls}, which this SDK version only
+supports as proto. The server predates proto response serialization for this type.
+```
+
+So the wire is not negotiable, and a JSON-only server fails at the first
+training step regardless of how correct its numbers are. Details worth knowing
+if you touch `proto_wire.py`:
+
+* Sample *requests* are still JSON. Only the two response types above, and the
+  `forward_backward` request, are proto.
+* `forward` is folded into the `forward_backward` endpoint upstream, carried by
+  a `forward_only` flag on the request.
+* The `Tensor` oneof is named `encoding` (`dense` / `sparse_csr`), not
+  `payload`.
+* `EncodedTextChunk.tokens` is **int32** while request tensors are float32 or
+  int64. `BatchedTensor.offsets` are int64 **byte** offsets with n+1 entries.
+* `proto_compress_fwdbwd` (zstd) is server-advertised and defaults off, so it
+  can be left off. The older `proto_write_fwdbwd` flag is no longer consulted.
+* `encode_forward_backward_output` packs only the fields present on *every*
+  datum, because a field missing from one datum would shift every later datum's
+  slice.
+
+[`test_proto_wire.py`](../../../tests/integrations/tinker/test_proto_wire.py)
+uses the SDK's own `request_conv` / `response_conv` converters as the oracle in
+both directions. That matters: comparing the router against hand-copied
+schema types cannot catch upstream drift, which is exactly how the JSON-only
+assumption survived as long as it did.
+
+## 6. Two conventions that silently produce wrong numbers
+
+Both of these fail by returning a plausible-looking number rather than by
+raising, which is why they get their own section.
+
+### Row alignment
+
+The router lays a row out as `[pad… prompt][response pad…]` — real tokens
+contiguous in the middle. Cortex's packer requires them in the *leading*
+columns. Aligning is not enough on its own: the log-probs come back in the
+aligned frame while the router's row slices index the original one, so
+`cortex.py` inverts the permutation on the way back. Aligning without inverting
+shifts every row's log-probs by its own prompt padding — no error, just a wrong
+number.
+[`test_skipping_the_inverse_would_shift_rows`](../../../tests/integrations/tinker/test_cortex_binder.py)
+is the discriminative test.
+
+### The final target token
+
+Tinker's datum convention is `model_input = full[:-1]`, `target_tokens =
+full[1:]`, with per-token `advantages` / `mask` / `logprobs` sliced `[1:]`. So
+`logprobs[k] = log p(full[k+1] | full[:k+1])` — the predicting-position frame,
+which matches Cortex's `compute_logprobs` exactly, with no shift needed.
+
+But `target_tokens[-1]` is a token that does **not** appear in `model_input`.
+Without appending it, the last scored position has nothing to predict: its
+log-prob is meaningless, and because the advantage sitting there still
+multiplies it, so is its gradient. The router appends it (kept out of
+`response_mask` and `advantages`, so it is scored *against*, never scored),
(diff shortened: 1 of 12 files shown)


sfc-gh-kganesan:
Pushed `cb54930`: `forward_backward_custom` now works, and `sample` refuses a temperature the trainer cannot reproduce.

**Custom loss.** The verb was unreachable rather than merely unsupported. It is client-side and hardcodes `cross_entropy` as its surrogate, which the router gated off, so it died on its first call. It now lowers to `weighted_logprob_sum`, and since Cortex registers no such loss the binder encodes it through stock `grpo` — the mechanism @sfc-gh-kganesan wrote in #100.

One improvement over #100 worth calling out: instead of sending π_old from a separate client forward, this **omits** `old_log_probs_shifted` entirely. grpo then defaults π_old to `logprobs.detach()` from the same forward, so the ratio is exactly 1.0 rather than 1 ± forward nondeterminism, and it costs no extra forward. Details in the PR body.

**Temperature.** Cortex has no `apply_temperature`, so the trainer scored at 1.0 whatever the sampler did — silent, and it biased the importance ratio. Now a 400, gated on a capability flag passed through `init_tinker_state` rather than a Cortex-specific check in the router, so on-prem does not inherit a limitation it may not have. This matters doubly for custom losses, where those same log-probs feed the client's autograd.

**Verification.** A sign or frame error here does not raise, it just trains the wrong way, so the tests are sized to that rather than to coverage:

- `TestGradientEquivalence` asserts `dL/dlogprobs == -weights` through the whole pipeline, with a negation-dropped control so it cannot pass vacuously.
- Live Cortex job: client-side CE `2.2032 → 0.0000` over 8 steps.
- Three rows of differing length each descend independently, which is the only way a per-row alignment shift surfaces. A single-row probe cannot catch it.

Also fixes a real bug in `forward`: the final target token was appended only for `forward_backward`, so pass 1's last log-prob scored a pad. Harmless for RL, silently wrong for a client differentiating those log-probs.

Limitations section in the body is updated — the remaining gaps are multimodal, `load_weights` (checkpoint resume), and `cispo`/`dro`. I also noted there that `ppo` and `importance_sampling` are indistinguishable on this path, since `actor_config` is dropped by `to_cortex_fwd_bwd_payload` and the ratio is always 1.

PR #111, opened 2026-09-14
Allow common registry imports without training extras

sfc-gh-mwyatt:
Summary:
  - Keep `arctic_platform.common` and `arctic_platform.common.registry` lightweight to import.
  - Preserve the dependency hint for lazy common exports.

  Testing:
  - `python -m unittest discover -s tests/common -p test_common_light_imports.py -v`
  - `python -m py_compile arctic_platform/common/__init__.py tests/common/test_common_light_imports.py`

PR #113, opened 2026-09-14
Align processor ABI and packed-loss apply (Phase A A0–A5)

### LINKED ISSUE #112: Plan A tracker: Cortex ↔ AP OSS split (zone === OSS)
Track **Plan A** for the Cortex ↔ AP OSS split. This issue is the **living checklist**. Dates, A-scaled calendar, and before/after architecture live in Neutrino `planning/cortex-ap-oss-exec-summary.md` (brief) and `planning/cortex-ap-oss-split.md` (gates). Do not treat the Cursor canvas or a Drive Doc as a second ledger — tick **here**.

**End state:** zone === OSS. Unify the training kernel in Arctic Platform. Keep the Neutrino/DSS **backend** proprietary (SnowAPI, zone, packed dispatch). On-prem backend stays AP HTTP/Ray.

**Staffing:** ~1.5 FTE. Clock for remaining dates: #113 merge assumed 2026-09-22. B3+C is **one** production image — do not split pin and cutover.

Phase A PR: #113 (open). A2: #111 (merged).

### Phase A — AP kernel (done 2026-09-20)

Landed behind `arctic-platform[sft]==0.1.3`. DSS still imports `arctic_training`. GSM8K 250s matched Cortex (TRL trainer 0.705 = 0.705; SkyRL pass@1 0.859 = 0.859).

- [x] **A0 ABI** — 5/4-arg callables; packing inner 6-arg + `pack=False`; `fwd_meta_keys`
- [x] **A1 hygiene** — public-name overwrite raises; frozen `PUBLIC_LOSS_FNS` / `PUBLIC_POST_PROCESSORS`
- [x] **A2 import** — #111
- [x] **A3 packed API** — `resolve_*` + `apply_packed_loss_reduction`; GRPO token/seq/prompt-mean on CPU
- [x] **A4 metrics** — `metric_is_summed`
- [x] **A5 compat** — zone names `grpo` / `compute_logprobs`

**#113 merge hygiene** (not A0–A5)

- [ ] Push Wyatt `238afb9` (`JobConfig.sp_size`, `summed_metrics=`, drop dead CCE check)
- [ ] Cortex shim sends `dp_size` with any global denom (#119 or fold into #113) — needed before the **zone pin**, not for the GSM8K 250s
- [ ] Merge #113

CPU bar: `pytest tests/rl tests/common tests/sft -m "not vllm and not gpu_serial"`

### Phase B — packaging (not started)

No `training.py`. Days, not the bake.

- [x] **B1** — AP `0.1.4` extras `[sft]`/`[rl]` actually import OPD/GRPO (clean-venv wheel, tag, index)
- [x] **B2** — `dss-platform` pin + `[gpu]` extra in one PR

Do **not** start until authorized. Do **not** bump the pin without B1 on the index.

### B3 + C — one bake (blocked on B1/B2)

Neutrino image + DSS `run_pipeline` cutover. Calendar **7 wk · 2026-11-10** (A-scaled **3 wk · 2026-10-15**). `pack=True` on DSS windows hangs SP/NCCL.

- [ ] **C1** — `_rl_fwd_bwd` calls AP `run_pipeline` with `batch`/`meta`
- [ ] **C2** — `_sft_fwd_bwd` still works when `processing is None`
- [ ] **C3** — one packed apply for DSS dispatch and on-prem packing
- [ ] **C4** — AT-dss `LOSS_FNS` removed or re-export-only
- [ ] **C5** — `causal_cross_entropy` / `compute_logprobs` / `metric_is_summed` in the **image**
- [ ] **B3 + B4 + C6** — Neutrino image; qa6 smoke (`NEUTRINO_TEST_DB.NEUTRINO_TEST_SCHEMA`); GRPO+SFT packed fwd_bwd; OPD `on_policy_distill` resolves in the image. **Zone === OSS.**

## Not this issue (cross-links)

These do **not** move the zone===OSS date. Do not wait on them to merge #113 or start B1/B2.

| Track | Where |
|---|---|
| BIRD / text2sql GS 128 MiB poll | [dss-client#61](https://github.com/Snowflake-AI-Research/dss-client/issues/61) |
| ZoRRo Train (SnowAPI + zone flag) | File a Cortex/DSS issue; AP `zorro_train` already exists as a patch |
| Capability parity (SP wrap + chunked LM head in AP) | File an AP issue **after C**. Extraction may overlap C; wrap cutover must not |

## Out of scope

- Second packed-apply site in AT-dss / zone
- `arctic_opd` or a second AP checkout
- Unifying Cortex `{args,kwargs}` vs on-prem `{batch,meta}` wire in the C PR
- Packing/SP topology as `build_model` checkboxes
- Raising the 128 MiB gRPC max; treating BIRD size 429s as retryable
- GPU-hours on SkyRL GSM8K 351 vs 529 length as a quality miss

sfc-gh-truwase:
## Summary
- Completes Plan A Phase A after #111: one 5/4-arg processor ABI, engine isolation via `_ENGINE_FWD_KEYS` + `fwd_meta_keys`, registry public-name hygiene, and `PackedLossReduction` resolve/apply in on-prem packing (`loss_only` + `scale_wrt_gas=False`).
- Lands the batch-dim contracts this ABI sits on: promote batch-dim tensors onto `batch` before DP split; `dp_size` is stamped as `num_workers` (disjoint shards; SP is validation-only); VeRL `rollout_is_weights` on `batch` only (non-tensor / `None` stays on `meta`).
- Registry is the AP∪Cortex union. Zone names **`grpo`** / **`compute_logprobs`** resolve unprefixed; AP-only logic is `ap_*`. Adds 5-arg `causal_cross_entropy` and `metric_is_summed` across packed, GAS, and DP combiners. Does not touch `common/__init__.py` (A2). Does not bump the DSS pin or cut over the zone.

Closes #112.

## Cortex vs on-prem Cortex-ABI parity (GSM8K 250)

On-prem HTTP on this branch, speaking the **Cortex catalog** (`loss_fn=grpo`, `post=compute_logprobs`, no client `dp_size`), matches Cortex zone quality on every **viable** arm. Recipe: Qwen3-1.7B, GSM8K, seed 42, 250 optimizer steps, 4 train + 4 sample GPUs. **Do not compare TRL numbers to SkyRL numbers** (n, seq, eval differ).

| Stack | Pair | Cortex vs on-prem+Cortex-ABI |
|---|---|---|
| **TRL** | trainer π_old | Quality match (all-run reward **0.705 = 0.705**). Length match (185 vs 187). |
| **TRL** | sampler π_old | Quality match within noise (all-run **0.720 vs 0.719**). Length match with current-token frame (last-100 130 vs 132). |
| **SkyRL** | trainer π_old | Quality match (held-out pass@1 **0.859 = 0.859**; last-100 reward 0.893 = 0.893). Length residual (last-20 351 vs 529) is leftover recipe (Liger+FA2 vs SDPA), not a quality miss. Pre-ABI `verl_grpo` on-prem was 0.663 / 0.711. |
| **SkyRL** | sampler π_old | Not viable (remote-engine `logprobs` ban). |
| **SkyRL** | client-side loss | Intra-Cortex only (≈ trainer). No on-prem client dump. |

TRL updates in these 250s are on-policy (`π_old ≡ π_new`): Cortex pops `old_log_probs`; zone `grpo` falls back to `logprobs.detach()`. Overlays that made on-prem speak this catalog live in the `cortex_training/` workspace, not this PR.

## Test plan
- [x] CPU gate: `pytest tests/rl tests/common tests/sft -m "not vllm and not gpu_serial"`
- [x] GitHub CI (HEAD `7434e4f`): formatting, unit-tests, Semgrep green
- [x] GSM8K 250 Cortex vs on-prem Cortex-ABI: TRL trainer + sampler quality match; SkyRL trainer pass@1 match (sampler not viable)
- [x] Engine isolation: `tests/rl/test_step0_contracts.py` (`test_engine_kwargs_drop_loss_tensors`, `test_fwd_meta_keys_can_add_but_not_blocked_keys`, packing/pack=False filters). `advantages` / `loss_mask` / `actor_config` never reach `engine()`
- [x] Packed apply: `tests/rl/test_phase_a_gates.py` TestA3 — GRPO token / sequence / prompt-mean, additive vs local, `n_mb=1` and `n_mb>1`, trio alignment, `scale_wrt_gas=False`
- [x] CCE: ignores `global_batch_size`; refuses missing `dp_size` when `batch_num_tokens` is set; context/config trio conflict raises; packed split-invariance
- [x] `dp_size` stamp: `_split_batch` stamps `num_workers`, not `world // SP`. Stamp-only `dp_size` does not scale token-mean; sequence-mean scales when `global_batch_size` is present. Live `dp_size` defaults to 1 and rejects non-positive
- [x] Metrics: `metric_is_summed` on packed, GAS (`combine_metric_microbatches`), and DP (`combine_metric_shards`); `{name}` vs `{name}.sum` mix raises; GRPO no longer emits mean `metrics["loss"]`
- [x] Names: catalog equality; overwrite raises; zone `grpo` / `compute_logprobs` resolve
- [x] Microbatch split: per-sequence `BATCH_DIM_CONTEXT_KEYS` only; unrelated length-`B` vectors are not split
- [x] Empty policy shard still runs ECHO on AP (`test_empty_policy_shard_still_runs_echo`); AT-dss source remains #167
- [x] No Phase B/C in this PR (no DSS pin, no Neutrino image, no zone `run_pipeline` swap)


arctic_platform/rl/processors/functional.py
```diff
@@ -59,29 +146,87 @@ def masked_normalization(
...
+            out[key] = config[key]
+    return out
+
+
 def agg_loss(
     loss_mat: torch.Tensor,
     loss_mask: torch.Tensor,
     loss_agg_mode: str = "token-mean",
-    dp_size: int = 1,
+    dp_size: Optional[int] = None,
```
sfc-gh-sbekman: why would a dp_size ever be None, it has to be at least 1, no?

arctic_platform/rl/processors/functional.py
```diff
@@ -198,9 +590,12 @@ def ppo_actor_loss_fn(
     cu_seqlens: torch.Tensor | None = None,
     loss_agg_mode: str = "token-mean",
     rollout_is_weights: torch.Tensor | None = None,
-    dp_size: int = 1,
+    dp_size: Optional[int] = None,
```
sfc-gh-sbekman: same as above and I see more of it below as well, but didn't flag

sfc-gh-sbekman (commented):
I skimmed though it - lots of code - in general looks ok - a few possible caveats below

arctic_platform/rl/processors/functional.py
sfc-gh-truwase: Agreed — a live `dp_size` is a DP width, so it is at least 1.

`agg_loss` / `ppo_actor_loss_fn` (and the other loss fns) now default to `dp_size: int = 1` and reject 0 / negatives. `None` is only the “key omitted from config” case inside `_resolve_dp_size`, so `batch_num_tokens` without an explicit factor still errors instead of silently using 1.

arctic_platform/rl/processors/functional.py
sfc-gh-truwase: Same change here: default `1`, never a missing scale.

sfc-gh-kganesan:
## Why on-prem and Cortex don't converge to the same place

Not the ABI. It's a loss-scale asymmetry that this PR's `dp_size` handling makes explicit.

### The asymmetry

Both paths use `loss_agg_mode="token-mean"` and both ship the same global denominator `batch_num_tokens = N`. Only the multiplier differs: on-prem sends `meta.dp_size = 4`; the Cortex shim strips it (`_cortex_shared.to_cortex_fwd_bwd_payload`, documented as deliberate).

Each rank computes `L_r = (1/N) * sum_{t in r} l_t * dp_size`, and DeepSpeed **averages** gradients over `R = 4` ranks:

- on-prem (`dp_size=4`): `g = (1/R) * sum_r grad L_r = grad S / N` — the true global token-mean
- Cortex (`dp_size=1`): `g = (1/R) * grad S / N` — **exactly 1/4 of on-prem, same direction, every step**

### Why that is usually invisible

Adam is invariant to a constant gradient rescale: scaling `g` by `c` scales `m_hat` by `c` and `sqrt(v_hat)` by `c`, which cancel whenever `eps << c*sqrt(v_hat)`. Decoupled AdamW decay never touches `g`. So a 4x loss scale should be a no-op.

What breaks the invariance is **global-norm gradient clipping**: `g -> g * min(1, tau/||g||)` is not homogeneous. Whenever `||g_cortex|| < tau < 4*||g_cortex||`, on-prem is rescaled and Cortex is not. The two then feed non-proportional gradient *sequences* into Adam and diverge through `v_hat`.

### Evidence: same 4x gap, two clipping regimes

| Stack | weight decay | grad clip | on-prem vs Cortex last-20 reward |
|---|---|---|---|
| TRL | 0.0 | **off** (absent from ds_config; DS default 0.0) | 0.723 vs 0.734 → **0.011** |
| SkyRL | 1e-2 | **1.0** (`ppo_base_config.yaml`) | 0.663 vs 0.889 → **0.226** |

TRL logs `grad_norm = 0.0` on all 250 steps in both arms (the Arctic client owns the optimizer, so TRL-side clipping never runs) → Adam absorbs the 4x. SkyRL clips at 1.0 → a 20x larger gap from the identical asymmetry.

### The trajectory is an overshoot, not an offset

SkyRL mean completion length by step window:

| steps | on-prem | Cortex trainer |
|---|---:|---:|
| 1–25 | 893 | 892 |
| 51–75 | 704 | 643 |
| 101–125 | **295** | 489 |
| 151–175 | **880** | 385 |
| 226–250 | 900 | 350 |

Identical through step ~75, then on-prem undershoots *below* Cortex's eventual equilibrium (295 vs ~350) and rebounds to the 1024 response cap, with reward going 0.772 → 0.482 → 0.663. Cortex descends monotonically and holds. That is the signature of too-large effective steps in a bistable length dynamic, which is what a 4x gradient predicts.

### Ruled out

- **Dropped π_old / clipping.** `max_abs_ratio_minus_1` is exactly `0.0` in all three SkyRL arms and ≤ 3.6e-3 in TRL, against `eps_clip=0.2`. The clip never binds, so π_old ≡ π_new shifts the per-token gradient by under 0.5%. (Worth noting separately: *neither* stack applies an importance correction despite `max_staleness_steps: 4`.)
- **Packed loss reduction.** `batch/microbatches_per_step: 1` in these runs, so `resolve_packed_loss_reduction` is a no-op. It only goes live at `n_mb > 1`.
- **A length-biased denominator.** The shim *does* forward `batch_num_tokens` from `meta`, so the per-token weight is `1/N` on both sides. This is a clean scalar, not a reweighting.

### Asks on this PR

1. Emit the **resolved** scale trio plus `loss_agg_mode` and `loss_is_additive` into step metrics. Today you cannot tell from a metrics dump which denominator a run used, which is why the 250-step compare can't be attributed.
2. Log the **pre-clip** gradient norm and the clip-hit rate. `grad_norm = 0.0` on every step is what makes this unfalsifiable from logs.
3. `_resolve_dp_size` (checked at head `e88c208`) correctly raises for `batch_num_tokens` without `dp_size`, and now rejects non-positive values. Please also reject `dp_size > 1` with **no** global denominator — that is the local-denominator × R case, and it still reaches `agg_loss` when a client passes `dp_size` through `processing.config`, since only the *stamped* value is dropped.
4. Publish `compute_logprobs` vs `ap_compute_logprobs` max-abs-diff on identical logits. The ~1e-3 kernel delta is the same order as the measured KL (6e-4), so the on-policy parity gate currently sits at the kernel noise floor.

### Single-variable test

Leave `dp_size=4` on the on-prem SkyRL run and change only `max_grad_norm` (raise or disable). If the gap collapses, the channel is clipping rather than the scale itself. The already-queued `SKYRL_DP_SIZE=1` alignment run tests the same hypothesis from the other side.

sfc-gh-kganesan:
## Suggested changes from the loss-scale analysis above

Concrete follow-ups to this PR, ordered by how much each one buys. All line refs are head `e88c208`. Nothing here blocks the ABI work — items 1–3 are what make an on-prem/Cortex compare attributable instead of guesswork.

### 1. Make the π_old fallback explicit rather than silent

`grpo.py:625-627`:

```python
old_log_probs_ctx = context.get("old_log_probs_shifted")
input_data = {
    "old_log_probs": old_log_probs_ctx.to(logprobs.device) if old_log_probs_ctx is not None else logprobs.detach(),
```

A missing alias sets π_old ≡ π_new, which makes `ratio` identically 1 and disables `eps_clip` entirely — silently, on both `ap_grpo` and `grpo`. Right now "the clip never bound" and "the clip was disabled by construction" are indistinguishable from any metrics dump, which is why attributing the 250-step gap needed three separate cross-checks.

Suggest:
- a `processing.config` flag (e.g. `allow_rederived_old_logprobs`, default `False`) and raise when π_old is absent and the flag is unset;
- emit `grpo/old_logprobs_source` = `provided` | `rederived` in metrics.

The docstring notes this is sound for one optimizer step per collected batch. That is a caller property, not a server one — SkyRL's `ppo_base_config.yaml` ships `max_staleness_steps: 4`, so the precondition isn't guaranteed by the callers this registry now serves.

### 2. Report the resolved loss scale

`_merge_distributed_config` (`grpo.py:845`, config wins) and `resolve_global_loss_scale` (`functional.py:172`, context wins) each resolve a trio that never reaches the response. Add to the metrics dict (`grpo.py:425`, plus the early-return dicts at `281` and `314` so it is present on every path):

`loss_scale/dp_size`, `loss_scale/batch_num_tokens`, `loss_scale/global_batch_size`, `loss_scale/loss_agg_mode`, `loss_scale/is_additive`

These are replicated across microbatches and shards, so the existing weighted-mean combine returns them unchanged — no `metric_is_summed` entry required.

This is the number that separated the TRL and SkyRL outcomes, and it is currently unrecoverable from a metrics dump.

### 3. Reject `dp_size > 1` with no global denominator

`_resolve_dp_size` (`functional.py:150`) receives only `batch_num_tokens`, so it cannot see the sequence-mean case, and it still permits the inverse hazard: `dp_size=4` against a **local** denominator, which multiplies a local token-mean by R.

Suggest passing `loss_agg_mode` and `global_batch_size` in and raising when `dp_size > 1` and that mode's global denominator is `None`, excluding `prompt-mean` (which intentionally does not multiply by `dp_size`).

This is reachable today: `dp_size` is in `_GRPO_CONFIG_KEYS`, and both merge paths drop only the *stamped* value, not a config-supplied one.

### 4. One source of truth for the default `loss_fn`

`pipeline.py:387` and `packed_reduction.py:137` each hard-code `"ap_grpo"` as the fallback. Two copies of a default that now selects between two different precedence rules is a drift hazard, and on a Cortex zone the executed contract should be `grpo`. Suggest a single shared constant, and consider requiring `processing.loss_fn` explicitly (raise on omission) now that the two contracts are not interchangeable.

### 5. Publish the log-prob kernel noise floor

`tests/rl/test_phase_a_gates.py:301-303` asserts the two kernels are distinct objects but never compares their output. Add a test that runs `compute_logprobs` and `ap_compute_logprobs` on identical logits/labels and asserts max-abs-diff below a recorded constant, with that constant visible in the test.

The delta is ~1e-3, the same order as measured mean KL (6e-4) and max |ratio−1| (1.2e-3 on-prem, 3.6e-3 Cortex). Any on-policy parity gate tighter than the kernel floor is measuring the kernel, not the training.

### 6. Assert temperature is applied exactly once

`temperature` is now on `_ENGINE_FWD_KEYS` (`pipeline.py:98`) while `apply_temperature` (`pipeline.py:888`) also divides logits, and `causal_cross_entropy._validate_neutral_temperature` refuses anything but 1.0. Add a test that temperature is applied once when both the engine kwarg and the post are present, or that the combination raises.

### Out of scope here

Pre-clip gradient norm and clip-hit-rate reporting belong to the trainer/optimizer side, not this registry — but that is the observable that decides whether a loss-scale asymmetry changes an outcome at all. Worth a separate issue.

sfc-gh-kganesan:
Pushed the DP-scale fix from the analysis above as a draft PR against this branch: #119 (5 files, +150/-10).

Scoped to the loss-scale issue only. Forwards `dp_size` in the Cortex shim, adds the mirror guard to `_resolve_dp_size` (`dp_size > 1` with a rank-local denominator, the complement of the guard you added in e88c208), and emits the resolved `(dp_size, batch_num_tokens, global_batch_size)` trio in GRPO metrics. CPU gate is green at 322 passed.

The regression test worth looking at first is `test_token_mean_is_dp_split_invariant` — full batch on one rank vs 4 uneven shards must agree exactly, and the `dp_size`-dropped variant lands at exactly `1/dp_size`.

One caveat carried over into the PR body: correcting this should make Cortex's reward *drop* toward on-prem's collapse, so it probably wants to land together with an LR reduction rather than alone.

sfc-gh-truwase:
@sfc-gh-kganesan — follow-up on the suggested changes, from the GSM8K 250s that matched Cortex by speaking this PR’s catalog (`grpo` / `compute_logprobs`, no client `dp_size`). Direction was **on-prem → Cortex as 113+zone exist today**, not changing Cortex to match old on-prem.

## In scope for 113 (keep / already landed)

Bare `grpo` / `compute_logprobs` (`554f989`), reject conflicting client `dp_size` (`e88c208`), SkyRL `None` batch-dim keys (`bab00ac`), `return_fwd_batch` (`4427f97`), skip zero-norm `engine.step` (`7434e4f`). Those are what the matched runs imported.

1. **Explicit π_old fallback** (`allow_rederived_old_logprobs` + `grpo/old_logprobs_source`): worth landing on 113 if you want one more commit. Silent `logprobs.detach()` at `grpo.py:625-627` is the contract TRL Cortex already uses (pops `old_log_probs`) and that SkyRL only hits if `old_log_probs_shifted` is aliased. A metric would have made that obvious. It does not change the quality match once both sides use the same fallback.
2. **Loss-scale trio on GRPO metrics**: cheap, belongs on this PR if you still want dumps to show `dp_size` / `batch_num_tokens` / `global_batch_size`. Attribution today is from code, not metrics.

## Out of scope for 113 (do not merge for this parity)

**#119** (forward `dp_size` on the Cortex shim). Not in the 250s. On-prem still stamps 4; Cortex still omits it; TRL all-run **0.705=0.705** and SkyRL pass@1 **0.859=0.859**. #119 would change Cortex, i.e. the opposite of the match. Rebase/follow-up, not a 113 merge requirement.

Client `dp_size=1` as a mimic: 113 already 500s (`context=4 config=1`). Correct client is **omit** `dp_size`.

3. **Reject `dp_size>1` with no global denominator**: good guard, did not fire on these runs.
4. **Single hardcoded `loss_fn` default**: the bug was `verl_grpo` vs `grpo` on the wire, not `pipeline.py` vs `packed_reduction.py` falling back to `ap_grpo`.
5. **Kernel max-abs-diff test**: hygiene; ~1e-3 is below TRL last-100 se ~0.024.
6. **Temperature-once**: T=1 on this recipe; Cortex already strips `apply_temperature`.

Clip / `max_grad_norm` A/B: trainer-side. Pre-ABI SkyRL collapse went away after the catalog overlay, without changing clip or landing #119.

Please keep 113 as the ABI + the five commits above. Observability (1)(2) optional on this PR; #119 and scale-into-the-zone stay a separate PR.

arctic_platform/common/utils/batch.py
```diff
@@ -22,6 +22,72 @@
...
+def _record_sp_candidate(candidates: dict[str, int], label: str, container: Any, key: str) -> None:
+    if not isinstance(container, dict) or key not in container:
+        return
+    value = container[key]
+    if value is None:
+        return
+    candidates[label] = resolve_parallelism_degree(value, key)
+
+
+def sp_size_from_job_config(job_config: Any) -> int:
```
sfc-gh-mwyatt, line 52: I feel that the `job_config` should be a single source of truth for jobs we create. It's validated as a pydantic model, so this sort of logic should ideally live as a validator on the config model itself. In general, this implementation feels overly defensive and could be easily replaced by something in the config model like:

```python
sp_size: int = Field(1, ge=1, validation_alias=AliasChoices(
    AliasPath("training_config", "sp_size"),
    AliasPath("ds_config", "sequence_parallel_size"),
    AliasPath("ds_worker_config", "sequence_parallel"),
))
```

arctic_platform/rl/processors/grpo.py
```diff
@@ -381,14 +695,248 @@ def grpo_loss(
...
+        "dp_size",
+        "batch_num_tokens",
+        "global_batch_size",
+        "entropy_coeff",
+        "use_kl_loss",
+        "kl_loss_coef",
+        "kl_loss_type",
+    }
+)
+_ECHO_REQUIRED_CONFIG_KEYS = frozenset({"aux_ce_weight", "echo_global_num_sequences"})
```
sfc-gh-mwyatt, line 745: This would likely be a lot easier to maintain and reason about if it was just a pydantic model that validates the config inputs. This feels like re-inventing parts of the pydantic validation - so let's just use a pydantic model!

sfc-gh-kganesan:
Followed the DP thread to a conclusion, and it turned into something more concrete than the quality argument I raised earlier. **This PR ships a guard that its own client violates.** Reproduced on CPU against pristine `e88c208`, no zone access needed.

`_resolve_dp_size` doesn't exist on `main` — d2ae536 introduces it here, with `batch_num_tokens requires an explicit dp_size`. The shim in this same PR forwards `batch_num_tokens` and `global_batch_size` from meta while omitting `dp_size`, which is precisely the rejected combination. Round-tripping the SkyRL Cortex payload:

```
client meta            : {'dp_size': 4, 'batch_num_tokens': 80, 'global_batch_size': 8}
wire processing.config : {'dp_size': None, 'batch_num_tokens': 80, 'global_batch_size': 8}
RESULT: ValueError -> batch_num_tokens requires an explicit dp_size
```

It's latent only because the zone runs a pinned image built from `main`, where `agg_loss` does `(sum / batch_num_tokens) * dp_size` with `dp_size` defaulting to 1 against a step-global denominator — silently `1/dp_size` of the global token-mean. It becomes a hard failure the moment the zone runs this PR's code, which is where kernel unification is heading.

On-prem was never exposed because `_split_batch` stamps `meta["dp_size"] = len(workers)` server-side, and only `http_server.py` / `ray_server.py` call it. `_cortex_dispatch` is a single remote call that never reaches that fan-out, and the Cortex wire `context` carries only `input_ids`, `advantages`, `loss_mask` — no meta bag. So `processing.config` is the only route to the zone, and it was being stripped. Structural asymmetry, not a config slip.

Fix is up as #119 against this branch, with the round-trip as a regression test that raises on `e88c208` and passes with the fix. CPU gate 323 passed.

**And a correction to my earlier comment.** I argued the 4x explained the quality gap and warned that fixing it would drag Cortex down to on-prem's collapse. Your parity result makes me think that's wrong. On-prem was correctly scaled all along via the `_split_batch` stamp, so if routing on-prem onto the unified kernel closed the gap while the Cortex arm was still at `1/dp_size`, the scale factor wasn't the driver — the kernel was, and Phase A's story holds up independently. That makes #119 a correctness and liveness fix with no expected quality change, which is a much safer thing to land.

The one loose end is whether the pinned zone honors `processing.config.dp_size` once it's actually sent. Can't be determined from the repo. A single `MODE=smoke` run with the `loss_scale/*` metrics in #119 settles it: `dp_size` should read 4 and `loss` should move by exactly 4x. If the metric reads 4 and the loss doesn't budge, the zone discards the key and needs its own change — worth knowing before Phase B.

Happy to split the three metric keys out as a standalone commit if you'd rather land observability first and keep this PR's diff where it is.

arctic_platform/common/utils/batch.py
```diff
@@ -250,6 +384,21 @@ def merge_dict_shards(shards_list: list[dict]) -> dict:
 
 _METRIC_PAIR_SUM_SUFFIX = ".sum"
 _METRIC_PAIR_TOKENS_SUFFIX = ".tokens"
+_SUMMED_METRIC_PREFIXES = ("loss_term_",)
+_SUMMED_METRIC_SUFFIXES = ("_sum", "_count")
+
+
+def metric_is_summed(key: str) -> bool:
```
sfc-gh-mwyatt, line 391: Seems like a very fragile way to indicate whether a metric should be summed. It might be more robust to add something when registering:

```python
@register_loss_fn("ap_grpo", summed_metrics=frozenset({"loss_term_rl", ...}))
```

arctic_platform/rl/processors/cortex_grpo.py
```diff
@@ -0,0 +1,84 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Cortex GRPO entrypoints: context-wins trio, registered as ``grpo`` / ``grpo_echo_v1``.
+
+Inner PPO/ECHO math is shared with ``ap_grpo``. The contracts differ: AP fills
+missing trio keys from meta/batch with config winning; Cortex
+``resolve_global_loss_scale`` lets context win and raises on a conflict.
```
sfc-gh-mwyatt, line 20: This contradiction in behavior seems problematic for switching between backends.

arctic_platform/rl/processors/causal_cross_entropy.py
```diff
@@ -0,0 +1,252 @@
...
+        or not isinstance(global_weight_sum, Real)
+        or not math.isfinite(float(global_weight_sum))
+        or float(global_weight_sum) < 0
+    ):
+        raise ValueError(
+            "causal_cross_entropy batch_num_tokens must be a finite non-negative "
+            f"global loss-weight sum, got {global_weight_sum!r}"
+        )
+    dp_size = _resolve_dp_size(dp_size, global_weight_sum)
+    if (
```
sfc-gh-mwyatt, line 113: Seems like this is overly defensive and unreachable given that `_resolve_dp_size` always returns an int!

sfc-gh-mwyatt (commented):
Additional thoughts:

There are many instances of added `value = config.get(field_name, default)` that I strongly dislike. This is unfortunately something that AI coding agents love to add as a defensive measure, but that I feel is a code smell. When we use `get` and define a default we are essentially collapsing three states into one: key absent, key present but None, and key present with a real value - which can make understanding how user-passed configs will be evaluated at runtime difficult to follow.

Additionally, it spreads defaults across many files and lines. I feel that user-provided configs and kwargs should be validated at input time and preferably by a pydantic model, where defaults, validation logic, etc. all live as part of a single class.

sfc-gh-truwase:
@sfc-gh-mwyatt on the review-level `config.get(field, default)` point: agreed on the principle and applied it where it was contained. GRPO defaults are now one table rather than ~25 call sites, and SP-size resolution moved onto `JobConfig` (validated at job create, sources scoped by `job_type`, no bool/float coerce).

I stopped short of pushing pydantic down the whole processor-config path in this PR. Those dicts come over the wire from four external clients and the kernel is already GPU-validated on GSM8K, so tightening unknown-key and coercion semantics here is the highest-risk change available. Most remaining `.get()` calls are presence probes on runtime `meta`/`batch` bags, where absence is the signal, not a defaulted config read.

Replies are on the five inline threads; the follow-ups are in `238afb9` / `d8b1f9d` / `cdb15f1`.

arctic_platform/common/utils/batch.py
sfc-gh-truwase: Agreed that the job config should own this — done. `JobConfig` now exposes an `sp_size` property and resolves it in a `model_validator(mode="after")`, so a bad or self-contradicting degree fails when the job is created rather than inside a worker. HTTP and Ray read `job_config.sp_size` directly.

I kept explicit conflict detection rather than `AliasChoices`, because `AliasChoices` takes the first matching source silently and we currently raise when VeRL's `ds_config.sequence_parallel_size` disagrees with `training_config.sp_size` — picking one arbitrarily would mis-size the DP world. It is a property rather than a `computed_field` so `model_dump()` stays the client wire shape for DeepSpeed workers. `sp_size_from_job_config` remains only as the adapter for that serialized dict.

Sources are also scoped by `job_type` (log-prob vs training) so a log-prob init that sends both `ds_config` and `log_prob_config` does not compare inactive blobs, and `resolve_parallelism_degree` rejects bool/float/string coercion.

arctic_platform/rl/processors/grpo.py
sfc-gh-truwase: I don't want to put a pydantic model on this path in this PR. These configs arrive as free-form dicts from VeRL, SkyRL, TRL, and the Cortex zone; `ap_grpo` / `grpo` are permissive about unknown keys while only `*_echo_v1` is strict. A model would make that uniform and reject keys those clients send today, and pydantic would coerce values that currently reach the math untouched — a numerics risk on a PR already validated by 250-step GSM8K runs.

What I did instead is collapse the frozenset and the scattered `config.get(key, default)` calls into one `key -> default` table per contract. The kwargs passed to the loss and the strict `*_echo_v1` schema are both derived from it, so they cannot drift. A test asserts the table covers every config-derived parameter of `_internal_grpo_loss_fn` and that the declared defaults match the signature defaults. That table is the seam for a real model once the ABI has settled and we can coordinate a wire-schema change with the clients.

arctic_platform/common/utils/batch.py
sfc-gh-truwase: Agreed — added. `register_loss_fn` now takes `summed_metrics=`, and the eight additive ECHO metrics are declared on `ap_grpo_echo_v1` / `grpo_echo_v1`. Declarations land in a flat union rather than per loss fn, because the three reducers run on metric dicts that have already been merged across microbatches, GAS, and DP ranks, so no loss-fn context survives to reduction time.

The naming convention stays as the fallback for undeclared keys, so nothing that is summed today can become a mean. A test asserts every declared name also still matches the fallback rule, which makes the migration a no-op for current emitters. Failed public-name overwrites bind first so they cannot leak names into that union.

arctic_platform/rl/processors/cortex_grpo.py
sfc-gh-truwase: I'd push back on this being a backend-switching hazard. The precedence is a property of the requested loss-fn name, not of the backend: `grpo` is the Cortex contract (context wins, conflict raises) and `ap_grpo` is the AP contract (config wins) on both backends — that name split is what the Cortex-zone entrypoints in this file exist for.

A client that keeps sending `grpo` gets identical scaling whether it hits the Cortex zone or this training kernel; a client only changes contracts by also renaming its loss fn. The module docstring now says that explicitly, and `TestTrioPrecedenceIsPerLossName` pins both precedences so unifying them later is a deliberate, visible change.

arctic_platform/rl/processors/causal_cross_entropy.py
sfc-gh-truwase: Correct — removed. `_resolve_dp_size` returns an `int >= 1` or raises, so every branch of that block was dead. The only user-visible effect is the error text for a bad `dp_size`, which now comes from `_resolve_dp_size` (`dp_size must be an integer >= 1, got ...`) instead of the CCE-specific message.

sfc-gh-mwyatt (approved):

PR #114, opened 2026-09-14
skyrl/cortex: self-installing launcher on the upstream tag

sfc-gh-kganesan:
Makes `simple_gsm8k_cortex` installable from a clean machine, and moves it onto
the upstream SkyRL tag.

## What changed

**`simple_gsm8k_cortex/run_qwen3_0.6b_gsm8k_grpo_cortex.sh`**

- Resolves dependencies with `uv run --isolated --extra skyrl-train --with <AP>[rl,cortex]`,
  replacing the conda env and `requirements.txt` setup. `skyrl` is built from
  `$SKYRL_HOME`, so the installed package matches the integration code on
  `PYTHONPATH`.
- Stops passing `generator.inference_engine.remote_urls`, which `skyrl-v0.3.0`
  rejects while parsing config. `external_server_urls` was already passed
  alongside it, so nothing replaces it. This removes the hand edit the recipe
  previously needed to run on the tag.
- Runs the `cortex_jobs` helper under `uv` as well. It runs *before* the
  training invocation, so it had no interpreter with `arctic_platform` on it and
  failed silently — which disabled release-on-exit and left a job holding GPUs.
- Sets `RAY_ENABLE_UV_RUN_RUNTIME_ENV=0`. Ray otherwise replays the driver's
  `uv run` line on each worker, which under `--isolated` fails with
  `Failed to spawn: --python`. Cortex owns the GPUs here, so the only Ray actors
  are driver-local and can share the driver interpreter.

**`simple_gsm8k_cortex/README.md`**

- Install section rewritten for the above.
- Corrects the description of what the launcher invokes.
- Replaces a troubleshooting row citing a `num_engines == len(remote_urls)`
  assertion that no longer exists at `v0.3.0`.

**`recipes/rl/skyrl/README.md`**

- Scopes the fork pin to the FSDP recipes. `named_non_persistent_buffers` is
  only reached by an FSDP worker's model construction, so the Cortex recipe can
  use the upstream tag. This also aligns with
  [cortex-training#16](https://github.com/snowflakedb/cortex-training/pull/16),
  which points at the tag.

## Why the previous install could not work

`requirements.txt` and `skyrl`'s own metadata pin different `flash-attn` wheel
URLs, and `uv --override` cannot arbitrate conflicting direct URLs — uv 0.12.1
and 0.12.15 both fail to resolve. Forcing past it floats `vllm` to `0.1.2`, and
`arctic-platform[rl]` omits `tenacity`, which the Cortex transport imports at
module scope.

Tracked in #115. This PR routes around those files rather than fixing them: on
the `uv run` path the flash-attn conflict does not arise, because SkyRL's
`[tool.uv.extra-build-variables]` sets `FLASH_ATTENTION_SKIP_CUDA_BUILD`, which
`uv pip install` never honored.

## How I tested

**Imports, clean environment.** Under the new invocation: `main_base`,
`arctic_platform.integrations.skyrl.entrypoint`, `_cortex_dispatch`, the Cortex
transport, `tenacity`, `ray` and `deepspeed` all import, with `skyrl` resolving
to `0.3.0` from the checkout.

**GSM8K against Cortex, full run on the upstream tag.** Before this PR I ran the
recipe to completion on `skyrl-v0.3.0` with the `remote_urls` line removed by
hand and deps from the old conda env: 233 steps, `eval/all/pass_at_1`
0.3033 → 0.7521 in 1h58m, against 0.2942 → 0.7680 in 2h03m on the fork pin. That
covers the tag move, which is the part of this PR that could affect results.

**GSM8K against Cortex, on the `uv run` path.** 20 steps, 0.2911 → 0.4193 →
0.5398, at 12–15s per generate versus ~30s on the conda env. Cut short by a
Cortex-side `504 Gateway Timeout` and API outage, unrelated to this change.

I did not repeat the full 233 steps on the `uv` path. It resolves newer
`torch`/`vllm`/`ray` than the conda env, but Cortex owns training and sampling —
generation goes through the inference-engine stub, so the driver's `vllm` is
never asked to generate, and no hyperparameter in the launcher changed. The
remaining delta is how deps are installed, which the 20 steps and the import
check exercise. Happy to run the full curve if a reviewer wants it on the record.

## Setup after this PR

```bash
pip install uv
git clone https://github.com/NovaSky-AI/SkyRL
git -C SkyRL checkout skyrl-v0.3.0
export SKYRL_HOME=$PWD/SkyRL
```

Then the `ARCTIC_CORTEX_*` variables, the dataset step, and the launcher.


sfc-gh-kganesan:
Scope correction on the fork pin, checked at `skyrl-v0.3.0`.

`named_non_persistent_buffers` sits in the `meta_init` branch of
`skyrl/backends/skyrl_train/workers/model_wrapper.py`, which only an FSDP
worker's model construction reaches. `ArcticRLExp._setup_trainer` builds
`ArcticPPOTrainer` against an inference-engine stub, so neither the ARL nor the
Cortex path constructs a local FSDP worker. The affected scope is the
`fsdp_*_entry.py` A/B baselines in `txt2sql/` and `long_context_qa/`.

The pin stays for the FSDP recipes, but because the sha is unreachable from any
upstream ref and is what those recipes were validated against — not because the
tag breaks them. Now reflected in the README.

sfc-gh-truwase (approved):

PR #118, opened 2026-09-17
client: resolve the Cortex connection from the cortex-training CLI's config

sfc-gh-kganesan:
## What changed

Running a Cortex recipe meant exporting `ARCTIC_CORTEX_HOST`, `_DATABASE`,
`_SCHEMA`, and `_PAT` by hand, even for users who had already run
`cortex-training login` and so had those same four values sitting in a file on
disk. `CortexConfig` now falls back to that file.

Resolution order, first hit winning:

1. Values passed explicitly, or present as `ARCTIC_CORTEX_*`.
2. `CORTEX_TRAINING_CONFIG`, naming a connection file without logging in.
3. The `cortex-training login` state, which records a config *path* rather
   than the config, so this is a double hop through
   `~/.config/cortex-training/login.json` (`XDG_CONFIG_HOME` respected).

`host`, `base_url`, `database`, `schema`, `endpoint`, and `pat` are read from
the file, and `CORTEX_TRAINING_PAT` covers a file that deliberately omits the
token, which that CLI documents.

This sits in `CortexConfig` rather than in the launcher on purpose: it is the
class every Cortex entry point constructs with no arguments, so one change
covers all of them instead of each launcher reimplementing it in bash.

## Two decisions worth reviewing

**The gate is `host`/`base_url`, not "any field set".** Those two select the
account and are mutually exclusive, so a file's `base_url` joining an exported
`host` would bypass PAT auth against a different target — the fallback is
skipped entirely when either is present. The remaining fields do fill in
individually, and an exported one overrides the file's copy, so exporting only
`ARCTIC_CORTEX_DATABASE` means the file's account with that database. Both
halves of that rule have tests.

**An unusable file reads as absent, with a warning.** A corrupt file, or a
login pointing at one that moved, yields no connection rather than an
exception, so the existing `cortex: set base_url (direct URL) or host (PAT
auth)` still names what the user has to fix. It logs a warning naming the
path, so a login gone stale is visible without being fatal.

It is implemented as a `mode="before"` validator rather than a custom settings
source because a source cannot see what higher-priority sources produced,
which is exactly what the merge rule above depends on.

## Launcher

The GSM8K launcher resolves the connection through the client up front and
prints the account it landed on:

```
Cortex target: <account>.snowflakecomputing.com NEUTRINO_DB.PUBLIC
```

It previously did not check at all. Both places that touch the connection
later — the pre-run job snapshot and the trainer — keep their errors
non-fatal, so an unset connection surfaced several minutes in as a Hydra
traceback. Now it exits immediately naming both options.

## How I tested

Thirteen unit tests next to the existing `TestCortexConfigReadsEnv`, covering:
the file supplying the connection via both `CORTEX_TRAINING_CONFIG` and login
state; the environment and explicit values each winning; a `base_url` file not
hijacking an exported host; a lone exported field overriding the file without
disabling it; a PAT taken from `CORTEX_TRAINING_PAT`; corrupt, non-object,
moved, and absent files each falling through to the existing error, warning
where a file was found; the file-sourced PAT still masked in `repr` and
`model_dump_json`; and an on-prem backend ignoring a present login. All 93
tests in `tests/client` pass.

The autouse env-isolation fixture now also clears `CORTEX_TRAINING_*` and
points `XDG_CONFIG_HOME` at an empty directory. Without that, a developer's
real login would satisfy configs that several existing tests expect to be
incomplete — those tests would have passed in CI and failed locally.

The launcher preflight was exercised end to end against a real account: login
state alone resolves and prints the target, exported variables take
precedence, `CORTEX_TRAINING_CONFIG` pointing at a `base_url` file reports
that URL, and an empty config home exits 1 with the error above.

Not rerun: a full GSM8K convergence job. This changes only how four strings
reach the client, and the previously verified path — exported variables — is
case 1, where the new code does nothing.


sfc-gh-kganesan:
Closing for now to keep focus on the cortex-training docs PR. The branch `karthik/skyrl-cortex-login` keeps the work: `CortexConfig` falling back to the `cortex-training` CLI's connection file, with tests, so `cortex-training login` alone configures a Cortex run. Worth reopening if we want to drop the `ARCTIC_CORTEX_*` exports from the recipe instructions; the open question is whether Arctic Platform should depend on that CLI's login-state layout.

PR #119, opened 2026-09-18
Forward dp_size to the zone and reject it without a global denominator

sfc-gh-kganesan:
Draft, targeting `sfc-gh-truwase/phase-a-abi-packing` rather than `main`. Scoped to the DP loss-scale issue only — no π_old, temperature, kernel-parity, or registry-default changes.

## The bug

`agg_loss` token-mean computes, per rank, `(1/N) * sum_{t in r} l_t * dp_size`, where `N` is the **global** token count. DeepSpeed then **averages** gradients across the `R` DP ranks, and the `* dp_size` exists precisely to cancel that `1/R`.

`_cortex_shared.to_cortex_fwd_bwd_payload` forwarded `batch_num_tokens` and `global_batch_size` but deliberately dropped `dp_size`:

- on-prem, `dp_size=4`: `g = (1/R) * sum_r grad L_r = grad S / N` — the true global token-mean
- Cortex, `dp_size` omitted → 1: `g = (1/R) * grad S / N` — exactly `1/R` of it

The gradient **direction is identical**; only the magnitude changes. So this presents as a 4x colder learning rate at 4-GPU DP, not as a failure, and nothing in the logged metrics distinguishes the two.

## Why it went unnoticed, and why it matters unevenly

Adam is invariant to a constant gradient rescale — scaling `g` by `c` scales `m_hat` and `sqrt(v_hat)` by `c`, which cancel. Decoupled AdamW decay never touches `g`. So the 4x should be a no-op, and in TRL it essentially was.

What breaks the invariance is global-norm gradient clipping, since `g -> g * min(1, tau/||g||)` is not homogeneous. The same asymmetry therefore lands very differently depending on the trainer's optimizer config:

| Stack | grad clip | on-prem vs Cortex last-20 reward |
|---|---|---|
| TRL | off | 0.723 vs 0.734 → 0.011 |
| SkyRL | 1.0 | 0.663 vs 0.889 → 0.226 |

## Heads-up: this fix will lower Cortex's numbers

Cortex has been the *better* arm in the 250-step runs (SkyRL: 0.889 vs 0.663 on-prem; held-out pass@1 ~0.86 vs 0.71). That is consistent with the bug rather than evidence against it — on-prem overshoots completion length down to 295 tokens, rebounds to the 1024 cap, and reward collapses from 0.772 to 0.482, while under-scaled Cortex descends monotonically to ~350 and holds.

Both Cortex arms agree closely despite computing the loss in completely different places (server-side 0.889/351 tokens, client-side 0.892/362), with on-prem the lone outlier — pointing at the shared shim rather than at loss placement.

So correcting the scale should make Cortex reproduce on-prem's collapse. **The conclusion is that on-prem is ~4x too hot for this recipe and Cortex was silently compensating**, which means this should land paired with an LR change (`lr=2.5e-7` is the obvious starting point) rather than as a standalone bugfix. Every on-prem-vs-Cortex quality comparison to date carries a 4x LR confound.

## Changes

1. `_cortex_shared.py` — forward `dp_size` alongside the denominators; correct the docstring that justified dropping it.
2. `_resolve_dp_size` — raise when `dp_size > 1` and the aggregation mode's global denominator is absent. This is the mirror of the guard already added in e88c208: that one catches a denominator with no factor, this one catches a factor with a rank-local denominator. `prompt-mean` is exempt because it never multiplies by `dp_size`. All four call sites updated.
3. `_grpo_loss` — emit `loss_scale/dp_size`, `loss_scale/batch_num_tokens`, `loss_scale/global_batch_size` (`-1.0` when absent, so a local denominator is distinguishable from a global one). Verified to pass through `combine_packed_metrics` / `combine_metric_shards` unchanged as replicated means.

## Test plan

- `tests/rl/test_losses.py::TestAggLoss::test_token_mean_is_dp_split_invariant` — the regression test for the whole class: one rank with the full batch vs 4 shards with deliberately uneven token counts must give the identical loss, and the `dp_size`-dropped variant must come out at exactly `1/dp_size`. CPU-only.
- `test_dp_size_above_one_requires_a_global_denominator` — both modes' denominators, `dp_size=1` exemption, `prompt-mean` exemption.
- `test_forwards_the_whole_global_loss_scale_trio` and `test_caller_config_still_wins_over_meta_for_the_scale_trio` in `tests/client/test_client_ops.py`.
- Updated `test_token_mean_and_dp_scaling`, which asserted `dp_size=2` against a rank-local denominator — the exact combination now rejected. Same expected value, now feeding `batch_num_tokens` as its own comment always said the caller should.
- Full CPU gate: `pytest tests/rl tests/common tests/sft tests/client -m "not vllm and not gpu_serial"` → 322 passed, 3 skipped, 14 deselected. Formatted with the repo's `black --preview --line-length 119` and `isort` settings.

## Not verified here

Whether the current DSS-pinned zone honors `processing.config.dp_size`. Your alignment doc records the zone as computing `* config.get("dp_size", 1)`, which implies it does and that no pin bump is needed — the first Cortex run with this branch confirms or refutes that, and a number that doesn't move means the fix needs the zone side too.


sfc-gh-kganesan:
Tracking issue with the standalone repro: #121

